PR #4, opened 2016-04-10
Update stdlib.ring

CalmoSoft:


MahmoudFayed:
Thanks Gal for adding the newlist() function.

PR #5, opened 2016-04-10
Update test.ring

CalmoSoft:


MahmoudFayed:
Thanks Gal
It would be better to set initial values for the variables a1 and a2 and avoid using the give command so the test can work more quickly (or without the need of input redirection).

PR #66, opened 2016-04-17
Update stdlib.ring

CalmoSoft:


MahmoudFayed:
Thanks for the new function!

PR #102, opened 2016-06-26
Ring Bitwise Operators( OR Operators )

Mr-alaa:
Ring Bitwise Operators( OR Operators )


MahmoudFayed:
Rejected, This branch has conflicts that must be resolve
Please use the last version of the file and don't modify the team file (i have already did that and changed the role from arabic movies to video tutorials).
So you need to update the resources file only but use the last version of this file.

PR #103, opened 2016-06-26
Ring Bitwise Operators Binary XOR 

Mr-alaa:
Ring Bitwise Operators Binary XOR


MahmoudFayed:
Rejected, because the updates are done on the ring-1.0-branch 
Please do the update to the master branch

PR #113, opened 2016-06-30
Update README.md

RibonsAmr:
The old command line wasn't working correctly.


MahmoudFayed:
Not correct because your removed the next libraries when you updated qt52
unixodbc mysql-connector-c allegro

PR #143, opened 2016-08-21
Update resources.data

ahmedhassouna:
New update of ring lectures


ahmedhassouna:
New update on ring lectures

PR #176, opened 2017-02-28
Create Draw-Circles-Dynamically-QPixMap2.ring

umariani:
Example of QPixMap2 for dynamic drawing on a canvas.


MahmoudFayed:
Thanks for  the sample

PR #184, opened 2017-03-31
Create AnalogClock

umariani:
Analog Clock - Ring
Use QPixMap2, Translate, Rotate, Scale


MahmoudFayed:
Thanks for the new sample

PR #187, opened 2017-04-13
RotateTextAndButton2.ring

umariani:
RotateTextAndButton2.ring
Added "Translate" function to the Class.
Displays 4 buttons with rotated text.


MahmoudFayed:
Thanks for the sample.

PR #195, opened 2017-04-21
Update rotatetextandbutton2.ring

umariani:
Added function  setButtonColor(color)


MahmoudFayed:
Thanks for the update

PR #214, opened 2017-05-21
FormDesigner - Fix adding custom horizontal headers for TableWidget

MajdiSobain:


MahmoudFayed:
Hello Majdi
Thanks for the update, I will accept it then I will do little changes to check for the Index
Because the user can change the Index start from 1 to 0.

PR #215, opened 2017-05-27
Update stdlibcore.ring

umariani:
Function:  Fsize(fh)  
Input:       filehandle
Returns:   file size
Name:      same format as  Fseek, Ftell, Fread  etc


MahmoudFayed:
Thanks for the new function.

PR #217, opened 2017-05-27
FormDesigner : Prevent crash at time of loading due to using (") at time of setting event actions

MajdiSobain:
I found this is easier than changing ( " ) into ( ' ) in ObjectDataAsString functions of all controllers


MahmoudFayed:
Hello Majdi

Thanks for the report. 

(1) A better way to apply your work around is using substr() 
Where you can replace the the '"' character with any other character

(2) We will not apply this fix, using this way or using substr() because we don't like to hide the bugs.

(3) The user must become able to write " " in properties, The use may need it in label captions for examples.

Greetings,
Mahmoud

MahmoudFayed:
Hello Majdi
I have already fixed the problem, Now we can use " in properties values
Keep up the good work :D 
Greetings,
Mahmoud

MajdiSobain:
Hello Eng. Mahmoud

You are welcome :)

I'm sorry for this prosthetic solution
But at least its used as a bug report ^_^

Thanks for the Fix :D 

Have a nice time

Greetings
Majdi

MahmoudFayed:
Hello Majdi

Thank you very much, Keep up the good work 

Greetings,
Mahmoud

PR #218, opened 2017-05-28
macOS System Integrity Protection (SIP) Fix

ribonsAMR:
System Integrity Protection (SIP) is a new protection system built-in El-Capitan and later mac operation systems: https://support.apple.com/en-us/HT204899
Due to the SIP we cannot install the ring inside /usr/ without disabling this protection system, instead we will use /usr/local/.

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -3,19 +3,19 @@
 
 # To be able to call ring from any folder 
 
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Hello Amr

Thanks for the update
Keep up the good work :D

Greetings,
Mahmoud

PR #219, opened 2017-05-28
RingQt : Correct column index in setItem() function of TableWidget class

MajdiSobain:


MahmoudFayed:
Thanks for the update

MajdiSobain:
You are welcome :)

PR #225, opened 2017-06-12
Create GetQuotesHistoryDraw-v1.2.ring

umariani:
GetQuotesHistoryDraw-v1.2.ring  --- Version 1.2
NEW updated version to handle changes in finance.yahoo.com
- Uses Ring Curl to get Cookies and Crumbs
- Yahoo uses new URL type interface, with epoch time and different parameters.
- Yahoo returns data in a different format

Please remove OLD vesion 1.0 and replace it with this version 1.2


MahmoudFayed:
Thanks for the update

PR #247, opened 2017-07-28
Create XML_VIN_Get.ring

umariani:
Using XML to get VIN - Vehicle Id Number - info from Web Site


MahmoudFayed:
Thanks for the sample

PR #250, opened 2017-07-31
Update CalmoSoftFifteenPuzzleGame.ring

CalmoSoft:
Please accept it.


MahmoudFayed:
Thanks for the update
Keep up the good work!

Note : Instead of using the constant : 1000
You can use : ClocksPerSecond() : http://ring-lang.sourceforge.net/doc1.4.1/dateandtime.html#clockspersecond-function

Because the value maybe changed between systems.

CalmoSoft:
Hello Mahmoud,

Thanks for your kind words.

Greetings,
Gal Zsolt
(~ CalmoSoft ~)

MahmoudFayed:
You are welcome :D

PR #256, opened 2017-08-08
Create Spiral-Animated.ring

umariani:
Spiral-Animated.ring
Draw animated spiral cycling through different colors.


MahmoudFayed:
Thanks for the new sample

PR #257, opened 2017-08-08
AA-Video-Music-Player-26.ring

umariani:
Version 26  ... Please replaces existing version  "AA-Video-Music-Player.ring
Fixed  pVolume()  non exist fuction
pVolume()  button is really a Label with an Icon
pVolume()  =  empty function


MahmoudFayed:
Thanks for the update

PR #276, opened 2017-08-25
Update calc-gui.ring

umariani:
New version of Calculator
Only 1 file  -- Calc-Gui.ring
Requires icons for math symbols


MahmoudFayed:
Thanks for the update

PR #278, opened 2017-08-25
Create GetQuotesHistoryDraw-v1.3.ring

umariani:
Version 1.3  GetQuotesHistoryDraw
Replaced the "Func WhereAreWe"
The jitteriness when the screen is made full size, when run under Windows is gone.
I do not have a MAC to be able to test it. (or adjust font sizes in a MAC(


MahmoudFayed:
Thanks for the update

PR #318, opened 2017-10-29
Implement autotools to ring

7snovic:
Implement the uinux-like autotools into ring, this is not a final commit, it's a kernel for upcoming commits.

TODO :- 
* set the prober path to move ring binary to.
* in ubuntu there are another program called `ring` comes with the `alliance` package so we need to make another check for that.
* add options to support --enable-**extension** in the ./configure phase instead of manually compile every extension.
* add a small shell file to handle the basic autotool steps, mainly to make sure of that autotools is exists on the system and execute the command `autoreconf -i`
* the previous shell script or the Makefile itself may structure a two important directories :- 
** /etc/ring/XXX to hold any upcoming probably configurations files.
** /usr/lib/ring/XXX to hold any *.so file in a one directory instead of being in the /usr/lib
* define some env variables.
* Upcoming TODOS.

until now, this files working grace in my side and compiled successfully, but still need some enhancements.

screenshot from 2017-10-29 23-17-45

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,0 +1,20 @@
+
+
+lib_LTLIBRARIES = libring.la
+libring_la_SOURCES = include/ring.h src/ring_state.c src/ring_ext.c \
+    src/ring_hashlib.c src/ring_hashtable.c src/ring_vmgc.c \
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Thanks, Keep up the good work!

PR #319, opened 2017-10-31
Autotools

7snovic:
This PR has 3 commits,

the first commit to implement the builder script,
the build scenario would be something like:
```
git clone ...
./build
./configure
make & sudo make install
```
-------

the second commit which create a default library directory to hold all of the Ring upcoming objects/shared libraries on the path : /usr/local/lib/ring/modules
modules may be trimmed on the future.

-------

the third commit to add a build directory and m4 directories to hold all the upcoming build m4 macros which will be added.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,19 @@
+#! /bin/sh
+
+## rights
+
+
+VERSION=1.4.5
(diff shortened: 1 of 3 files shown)


MahmoudFayed:
Thanks

PR #320, opened 2017-10-31
Pcre2

7snovic:
This will be a *long-term* PR and must still open until finishing this work,

What have been done so far:-
- populate the ring files.
- install the library source files.
- write the first function pcre_match
- write some tests
....

I will push the full TODO list soon, so any contributions will be welcomed.


MahmoudFayed:
Very nice idea, This will be an important extension.
Keep up the good work!

7snovic:
I guess that we can go with the extension right now,
the extension implement one method so far `pcre_match` with some available options like the case sensitive and global option.

I had defined a version constant `RING_PCRE2_VERSION` to track the upcoming changes and improvements in the future,
one of the most important TODOs is to implement pcre_study and use jit engine which I will implement in the upcoming versions ESA.

kindly review buildvc.bat file, there are two important flags that must be passed in the compilation phase to pcre library,
`HAVE_CONFIG_H` and `PCRE2_CODE_UNIT_WIDTH=8`

@MahmoudFayed

MahmoudFayed:
Thanks, Keep up the good work

PR #321, opened 2017-11-01
add changelog file

7snovic:
@MahmoudFayed we may need to fill up this file with the new changes in the core&language API.


MahmoudFayed:
We write a description for each commit 
Important information (New Features) are added to (What is new?) section in the documentation
Long change-log file with all of the details is not necessary for the project.

7snovic:
I see, but what I meant by such a change log file, is a file to hold the changes in the API itself, which I don't think that -from my point of view- `What is new?` will be the proper place to list such a changes.

Also in the future, when releasing more than one LTS version, and when we will have a multiple branches/versions to commit against, digging in commits will not be an efficient to track the feature/change.

this is just a POV but it's Okay at all to not having such a file.

7snovic:
Also it's important to mention that Changelog file is a part of the GNU standards

https://www.gnu.org/prep/standards/html_node/Change-Logs.html

MahmoudFayed:
1 - (What is new?) chapters are more readable (High quality documentation that cover every big of information is part of the project)
2 - We will not have more than one LTS version (We keep compatibility + Moving forward in one direction)
3 - Ring is not a GNU project, We don't have to follow the GNU standards
4 - Having Change-Log file + (Project Documentation + Code Commits Description) = unnecessary replication at the current stage of development, If we need this in the future, We may add it.

PR #323, opened 2017-11-11
Autotools

7snovic:
the RING.m4 is still under constructions, and can not be used right now.

this commit is part of autotools patch, this aims to give the ability to use autotools and make the benefits of it.

the upcoming is to give the user the power of using ./configure script to choose his preferable flavours of ring extensions to be compiled with ring.


MahmoudFayed:
Thanks for the update

PR #324, opened 2017-11-11
Hotfixes | fix readlink & chdir warnings

7snovic:
During compiling ring we got a lot of warnings -not an actual errors- like : 

> warning: ignoring return value of ‘readlink’, declared with attribute warn_unused_result [-Wunused-result]

this branch and commits aims to fix & git rid of this warnings.

However, the only change in ring_stmt.c file was in line [261](https://github.com/ring-lang/ring/blob/master/src/ring_stmt.c#L261), but it seems that was a lot of white-spaces that my editor trim it by default :

```
diff --git a/src/ring_stmt.c b/src/ring_stmt.c
index 598948a..fe5b261 100644
--- a/src/ring_stmt.c
+++ b/src/ring_stmt.c
@@ -258,10 +258,7 @@ int ring_parser_stmt ( Parser *pParser )
                        /* Read The File */
                        x = ring_scanner_readfile(pParser->pRingState,cFileName);
                        /* Restore the Current Directory */
-                       ring_chdir(cCurrentDir);
+                       if (ring_chdir(cCurrentDir)) {
+                               ring_parser_error(pParser,RING_PARSER_ERROR_DIRNAME);
+                               return 0;
+                       }
                        /*
                        **  Generate Code
                        **  Return NULL
```

### DIFF
--- a/include/ring_parser.h
+++ b/include/ring_parser.h
@@ -71,6 +71,7 @@ typedef struct Parser {
 #define RING_PARSER_ERROR_PARENTLIKESUBCLASS "Error (C24) : The parent class name is identical to the subclass name"
 #define RING_PARSER_ERROR_ACCESSSELFREF "Error (C25) : Trying to access the self reference after the object name"
 #define RING_PARSER_ERROR_CLASSREDEFINE "Error (C26) : Class redefinition, class is already defined!"
+#define RING_PARSER_ERROR_DIRNAME "Error (C27) : Error in directory name"
 /* Functions */
 
 int ring_parser_start ( List *pTokens,RingState *pRingState ) ;
--- a/include/ring_state.h
+++ b/include/ring_state.h
@@ -77,7 +77,7 @@ int ring_currentdir ( char *cDirPath ) ;
 
 int ring_exefilename ( char *cDirPath ) ;
 
-void ring_chdir ( const char *cDir ) ;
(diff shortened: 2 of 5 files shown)


MahmoudFayed:
Thanks for the update, but the C source code is generated using PWCT (http://pwct.org)
Don't modify the C source code directly!

(1) Download PWCT 1.9 (Art) From : https://sourceforge.net/projects/doublesvsoop/files/PWCT%20For%20MS-Windows/PWCT%201.9%20%28Art%29/Fayed_PWCT_1.9_Art.exe/download

(2) Modify the Visual Source Files in this folder : https://github.com/ring-lang/ring/tree/master/visualsrc
Then generate the C code.

I will accept this pull request and I will do the visual source updates myself, but In the future use PWCT and update the visual source files in your pull request.

7snovic:
OK, I will put that in mind.

MahmoudFayed:
Keep up the good work!

7snovic:
@MahmoudFayed it seems that I have some troubles with running PWCT. There are some problems with wine software in my environment [ubuntu 17], I guess that there are some alternatives solutions to complete fixing those errors & warnings,
1 - I fix it by myself and you update the visual source files.
2 - I log these warnings here and you fix it in the visual source code then generate the C source code.

Regards.

PR #433, opened 2017-12-19
BitCoin-BlockChain.ring

umariani:
BitCoin-BlockChain.ring

How to use Class, Variables, and Functions in an OOP's manner


MahmoudFayed:
Thanks

PR #469, opened 2017-12-24
added lesson 2

Rashadse:
I thought it would be more convenient if i tested the commands from lesson 2 in 1 file only, instead of having 7 test files overcrowding the repo.


MahmoudFayed:
Thanks

PR #491, opened 2017-12-30
lesson 11

darshadam:


MahmoudFayed:
I will accept it, But the sample "hidden rar in jpg" will be removed!

PR #562, opened 2018-02-04
Update foxring.ring

japfrosado:
Function frTransform() updated with "@X" format function and added the following functions frAsc(), frAtC() and frChr()


MahmoudFayed:
Thanks for the update

PR #576, opened 2018-02-18
Update sample.ring

japfrosado:
Test new functions


MahmoudFayed:
Thanks for the update

PR #580, opened 2018-03-01
Update foxring.ring

japfrosado:
frAsc() was updated.


MahmoudFayed:
Thanks for the update

PR #617, opened 2018-05-04
Update foxring.ring

japfrosado:
Added frProper() function


MahmoudFayed:
Thanks for the update

PR #645, opened 2018-07-11
Update sample.ring

japfrosado:


MahmoudFayed:
Thanks for the new sample!

PR #651, opened 2018-07-31
Create NASA-1-Fibonacci.ring

umariani:
NASA set of problems
1 - Fibonacci


MahmoudFayed:
Thanks for the new sample!

PR #666, opened 2018-09-06
Update QtClassConverterController.ring

MajdiSobain:
Correct three errors:
1- Detect virtual class destroyer.
2- Correct duplicated end brackets.
3- Avoid affecting enums\flags if some are substring of others.


MahmoudFayed:
Thanks for the update 
Keep up the good work

PR #668, opened 2018-09-07
Create FourBitAdder-Expands-to-256-Bits.ring

umariani:
FourBitAdder can expands to any number of Bits
Sample shows 4, 32, 64, 128, 256 Bits


MahmoudFayed:
Thanks

PR #670, opened 2018-09-07
Update oop.txt

umariani:
Example of using Class to Add 2 numbers.
Details showing Class:  Operator, Para, Type: Object, String, Number


MahmoudFayed:
Thanks

PR #699, opened 2018-10-12
Fix ring2exe.ring problem with spaces in paths

justgo97:
There is a problem with path resolving in ring2exe when the ring folder path have spaces for example :

`C:\Program Files\ring`

to fix this i just added two " around the ring path i don't know if this is the best way to do it but it works for me


MahmoudFayed:
Thanks for the update

PR #704, opened 2018-11-24
Update stdlibcore.ring

CalmoSoft:


MahmoudFayed:
Hello Gal
Very nice idea, but I think we can add a better function that support different formats with a better name
We can work on this in Ring Group
Thanks

PR #705, opened 2018-12-06
Change line 90

froottools:
Change;
sudo apt-get libtool install m4 automake
to
sudo apt-get install libtool m4 automake


MahmoudFayed:
Thanks for the update

PR #808, opened 2019-05-19
Update lists.txt

umariani:
Return Parameters from a Function in an Array or Hash Format


MahmoudFayed:
Thanks for the update

PR #839, opened 2020-02-10
Fixed syntax error (Using RingRayLib)

pancakevirus:
Hello.

The following markup position is wrong. Could you check it.

　
.. index:: 
	pair: Using RingRayLib; Functions


MahmoudFayed:
Thanks

PR #845, opened 2020-06-11
Update stdlibcore.ring

CalmoSoft:


MahmoudFayed:
Hello Gal
We will remove extra functions from stdlib that are not used too much in applications
Greetings,
Mahmoud

PR #886, opened 2020-08-30
Update applications/typingquiz

zaka7024:
More improvements and the user now can use the backspace key to correct the errors.


MahmoudFayed:
Thanks

PR #903, opened 2020-10-01
RingNotepad: Add debug and run console features in nonWindows OS

MajdiSobain:
Hello Mahmoud

These two features are now working as expected in Ubuntu but need to be tested in MacOS before being merged

I hope they are working good


MahmoudFayed:
Thanks

MahmoudFayed:
I will test them on macOS very soon
If they need an update, I will do it
Keep up the GREAT WORK

PR #907, opened 2020-11-24
Sum the apostrophe character as a highlight operator in Notepad++ Ring-Lang tool

bahasalien:
There we go & a few fixed typos of course.


MahmoudFayed:
Thanks

PR #910, opened 2020-12-13
Update Socket For Work on Multi Platform

abdallah-elsharif:


MahmoudFayed:
Thanks for the update

PR #911, opened 2020-12-13
Update Socket use buffer in recv function

abdallah-elsharif:


MahmoudFayed:
Thanks for the update

PR #914, opened 2020-12-17
Fix crash in ring_vm_numtostring when printing a large double value

idrassi:
The function `ring_vm_numtostring` uses `sprintf` on a fixed size `cStr` buffer with a length of 100 bytes.
But if the double value requires more than 99 characters to be printed, then the `sprintf` call will overrun the buffer cStr, causing a crash.

By using `snprintf` instead of sprintf, we avoid the crash since we can handle gracefully the special case of double value requiering more than 99 characters by using in this case the compact double presentation (%e).

Example of crashing program:
```
c = 999999999999999
for i = 1 to 13
  c *= 999999999999999
end

put "c = " + c + nl
```

Crash reproduced on Windows 32-bit and Ubuntu Linux 64-bit.
Below is screenshot of crash of the above program under Ubuntu Linux 64-bit:

ring_crash_linux_ubuntu


MahmoudFayed:
Thank you very much
Keep up the GREAT WORK

PR #917, opened 2020-12-19
Socket

abdallah-elsharif:


MahmoudFayed:
Yes, always use RING_API_GETSTRING2() and RING_API_RETSTRING2() to support binary data
Because Ring strings support that
Thanks for the update

PR #918, opened 2020-12-20
Speedup of hex2str implementation (10x) and better error code if input has odd number of chars

idrassi:
`hex2str` implementation was using `scanf` to convert from hexadecimal format which very slow. The new code uses simple hexadecimal conversion logic which is 10x faster.
`hex2str` implementation was expecting an even number of characters to process but didn't check for it. For example, if `nMax` is 1, the code will read past the end of the string at index 1. Luckily, the extra byte is zero and so a `scanf` error is displayed but it is better to check `nMax` from the beginning and report more adequate error.


MahmoudFayed:
Thanks for the update

PR #919, opened 2020-12-20
Fix inconsistency of hex() function between Linux-64bit, Linux-32bit and  Windows-64bit

idrassi:
`hex()` was returning different outputs for the same integer value on different platforms.
For example, output of `hex(999999999999999)` is "`38d7ea4c67fff`" on Linux-64bit but it is "`a4c67fff`" on Linux-32bit and Windows-64bit.
The issue was caused by the use of (`unsigned long`) type which has different size depending on the build platform.
The solution is to use (`unsigned long long`) which is guaranteed to be 64-bit on all platforms.

### DIFF
--- a/language/src/ring_api.c
+++ b/language/src/ring_api.c
@@ -1221,15 +1221,16 @@ void ring_vmlib_hex ( void *pPointer )
 
 void ring_vmlib_dec ( void *pPointer )
 {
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Thanks for the update

PR #920, opened 2020-12-21
Socket : Support UDP Protocol

abdallah-elsharif:


MahmoudFayed:
Thanks for the update

PR #921, opened 2020-12-21
Update Socket : update samples

abdallah-elsharif:


MahmoudFayed:
Thanks for the update

PR #922, opened 2020-12-21
Fix Number() inconsistency with String(), add better error handling, make String() better handle integers

idrassi:
`Number()` was not accepting strings created by `String()` if they are more than 15 characters. This made it impossible to load results from previous calculations. Moreover, `Number()` was not handling error cases correctly because of its use of `atof`.
The solution is to remove length limitation and to use `strtod` function which provides better error handling.

`String()` was printing decimal point even for exact integer values. This change makes it better handle exact integer values up to 53 bits in size ( `-9007199254740991` <= |value| <= `9007199254740991`)

Test program:

```
c1 = 999999999999999
for i = 1 to 13 {
	c1 *= 999999999999999
}

s = string(c1)
? "c1 = " + s
c2 = number(s)
? "c2 = " + c2

? "c2 - c1 = " + (c2 - c1)

str1 = "-2222044646462"
c = Number(str1)
str2 = String (c)

if str1 = str2 {
	? "Strings identical"

else
	? "Strings mismatch!!"
}
```


MahmoudFayed:
Thanks for the update
Keep up the GREAT WORK

PR #923, opened 2020-12-23
Fix memory leak in OpenSSL extension randbytes() and better handling of errors

idrassi:
in `ring_vm_openssl_randbytes`, the pointer `cStr` was never freed. It was also allocated with an extra byte without reason and there was no check if `malloc` failed or not.

Finally, if the call to OpenSSL `RAND_bytes` failed, `ring_vm_openssl_randbytes()` was returning an an integer equal to 0 but it should fail instead


MahmoudFayed:
Thanks for the update

PR #924, opened 2020-12-23
Fix memory leak re-introduced by commit 4b8f07fc9c1f7e2537101dd4f68593a01e964ebf "Update Visual Source (using PWCT)"

idrassi:
Unfortunately, during code refactoring of `ring_vm_openssl_randbytes`, the "`free`" statement was lost


MahmoudFayed:
Thanks for the update

PR #926, opened 2020-12-26
Speedup string deep-copy by allocating memory only when needed and using memcpy for long size

idrassi:
Existing implementation didn't perform fast enough in code that uses string variable with long and fixed size.
New code is 2x faster for size >= 1KiB and 3x faster for size >= 4KiB.
For example, the code below reuses the same string variable `s` to read 4KB from a file in order to hash it.
The previous code was allocating/freeing pointers unnecessary and also it was copying content byte by byte
```
f = fopen (fileName, "rb)
h = sha256init()
while true {
  s = fread(f, 4096)
  if isstring(s) {
    sha256update(h, s)
  else
    exit; // end of file reached
  }
}
digest = sha256final(h)
fclose (f)
```

### DIFF
--- a/language/src/ring_string.c
+++ b/language/src/ring_string.c
@@ -60,15 +60,23 @@ RING_API void ring_string_set2_gc ( void *pState,String *pString,const char *str
 	int x  ;
 	assert(pString != NULL);
 	assert(pString->cStr != NULL);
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Thanks for the update

PR #927, opened 2020-12-27
Speedup string append (10x to 800x) by using realloc instead of malloc/free and fix ring_state_realloc implementation

idrassi:
The code was using `malloc-copy-free` strategy which is not optimal since C runtime provides an optimized `realloc` implementation that gives huge speedup in scenarios where we need to append data to an existing pointer.
Moreover, the implementation of `ring_state_realloc` was not correct since it didn't handle the case of pointers belonging to the internal memory pool which can cause issues.

The example below demonstrates the performance gain: on Linux 64bit, before change it gives a `8.17` seconds performance and after the change it gives `0.01` seconds.

```
fullStr = "initial"
lineStr = "Dummy content line"

t1 = clock()
for i = 1 to 20000 {
	fullStr+= lineStr
}
t2 = clock()

delta = (t2 - t1) /clockspersecond()

? "total length = " + len(fullStr) + " characters"
? "perf= " + delta + " seconds"
```

### DIFF
--- a/language/include/ring_pooldata.h
+++ b/language/include/ring_pooldata.h
@@ -20,6 +20,7 @@ typedef struct PoolData {
 		HashItem vHashItem  ;
 		char cChar[RING_POOLMANAGER_ITEMSIZE]  ;
 	} PoolDataTypes ;
+	int nAllocatedSize;
 	struct PoolData *pNext  ;
 } PoolData ;
 typedef struct PoolManager {
--- a/language/include/ring_pooldata.h
+++ b/language/include/ring_pooldata.h
@@ -20,7 +20,6 @@ typedef struct PoolData {
 		HashItem vHashItem  ;
 		char cChar[RING_POOLMANAGER_ITEMSIZE]  ;
 	} PoolDataTypes ;
-	int nAllocatedSize;
 	struct PoolData *pNext  ;
 } PoolData ;
 typedef struct PoolManager {
--- a/language/include/ring_vmgc.h
+++ b/language/include/ring_vmgc.h
@@ -57,6 +57,8 @@ void ring_poolmanager_newblock ( RingState *pRingState ) ;
 
 void * ring_poolmanager_allocate ( RingState *pRingState,size_t size ) ;
 
+int ring_poolmanager_find ( RingState *pRingState,void *pMemory ) ;
+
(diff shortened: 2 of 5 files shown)


MahmoudFayed:
[1]  Thanks for this pull request

(1.1) Very nice update, notable increase in the performance and it passes all of the tests (Including my manual tests too) - So we can merge it without problems.

(1.2) In this performance test : https://github.com/ring-lang/ring/blob/master/language/tests/scripts/performance/strcatperformance.ring
On my machine, the time is changed from (0.60 second to 0.01) which means 60x Faster! (Really GREAT)

(1.3) By Testing the performance of Ring Code Generator for Extensions, when used with RingQt 
On my machine, the time is changed from 73 seconds to 18 seconds which means 4x Faster in a real application (VERY NICE)

[2] Little comment about the implementation & updates

(2.1) Updating the structure (PoolData) and adding (int nAllocatedSize) is not the right thing to do with respect to the (Memory Usage) because each allocated item will have an extra size (The size of the integer) - And since we allocate 2 millions items in the start of the program - This means around (8 MB)

(2.2) Using (int nAllocatedSize) added little complexity to the solution where you have to be sure that it contains the correct value and keep modifying it from many functions.

[3] Suggested Solution

(3.1) Remove (int nAllocatedSize) from the PoolData structure
(3.2) Update ring_state_realloc() to get the (old size) as an extra parameter before the new size
To be like this
RING_API void * ring_state_realloc ( void *pState,void *ptr, size_t nAllocatedSize, size_t size )
(3.3) Modify ring_string_add2_gc() to pass the current string size when using ring_state_realloc() 

This way we get (Better Performance) + (Without Increasing the Memory Usage)

Keep up the GREAT WORK

idrassi:
Thank you for this comprehensive feedback. I'm glad the speedup is confirmed in real world scenarios.

I agree with your suggested solution. I will update the pull request with a commit that implements it.

MahmoudFayed:
You are welcome

idrassi:
I have pushed the extra commit for the suggested solution and it contains also a modification of ring_string_new2_gc to use memcpy for values larger than 64 bytes for speedup purposes.

MahmoudFayed:
Thanks for the update
Keep up the GREAT WORK

PR #928, opened 2020-12-28
update ring_vmfile.c : add listdir function

abdallah-elsharif:
listdir function only gets exist folders and files in the path which passed as a parameter
without gets sub folders content


MahmoudFayed:
Thanks for the pull request

Just a little update
Before using RING_API_ERROR
Free the allocated string (str)
Because the Ring Code may call your function inside (Try/Catch/Done)
And avoid the runtime error
This will lead to a memory leak because of (str) is not deleted

Keep up the GREAT WORK

PR #929, opened 2020-12-29
Fix inconsistency of fexists between Linux/Windows and add functions direxists and getpathtype

idrassi:
Under Linux, `fopen` succeeds even if the input is a directory and not a file and so `fopen` can't be used to check if input is for an existing file name or not.
On the other hand, `stat` function provides a cross platform way to detect if the given path exists and also to get its type (file, directory...). So, in this implementation we use `stat`.

Before, there was no easy way to get the type of a give path, check if it is valid or not,  or check the existence of a directory.
The two functions `direxists` and `getpathtype` solve this by adding this functionality.

**Direxists**
We can check if a directory exists using the `Direxists() `function

Syntax:
`Direxists(cDirPath)` ---> returns 1 if the directory exists

Example:
```
see direxists("b:\mahmoud\apps\ring") + nl +
    direxists("b:\mahmoud\apps\ring\bin2") + nl
```

Output:
```
1
0
```

**Getpathtype**
We can get the type a given path (file or directory) using the `Getpathtype() `function

Syntax:
`Getpathtype(cPath)` ---> returns the following value:

- 0 if the path doesn't exists
- 1 if it corresponds to existing file
- 2 if it corresponds to existing directory
- -1 if the path exists but has an unknown type (e.g. a pipe)

Example:

```
see Getpathtype("b:\mahmoud\apps\ring\ring.exe") + nl +
    Getpathtype("b:\mahmoud\apps\ring") + nl +
    Getpathtype("b:\mahmoud\apps\ring\bin2") + nl
```

Output:
```
1
2
0
```

### DIFF
--- a/language/include/ring_state.h
+++ b/language/include/ring_state.h
@@ -97,6 +97,10 @@ RING_API void ring_state_log ( RingState *pRingState,const char *cStr ) ;
 
 int ring_fexists ( const char *cFileName ) ;
 
+int ring_direxists ( const char *cDirPath ) ;
+
+int ring_getpathtype ( const char *cPath ) ;
+
 int ring_currentdir ( char *cDirPath ) ;
 
 int ring_exefilename ( char *cDirPath ) ;
--- a/language/include/ring_state.h
+++ b/language/include/ring_state.h
@@ -97,9 +97,7 @@ RING_API void ring_state_log ( RingState *pRingState,const char *cStr ) ;
 
 int ring_fexists ( const char *cFileName ) ;
 
-int ring_direxists ( const char *cDirPath ) ;
-
-int ring_getpathtype ( const char *cPath ) ;
+int ring_fexists_general ( const char *cFileName ) ;
 
 int ring_currentdir ( char *cDirPath ) ;
 
--- a/language/include/ring_state.h
+++ b/language/include/ring_state.h
@@ -97,8 +97,6 @@ RING_API void ring_state_log ( RingState *pRingState,const char *cStr ) ;
 
 int ring_fexists ( const char *cFileName ) ;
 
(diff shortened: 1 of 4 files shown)


MahmoudFayed:
Thank you very much

Suggestions: 

(1) Don't modify ring_state.h  and ring_state.c ---> Keep them as before including ring_fexists() implementation

The reasons behind that 
1.1 We have a specific architecture of different layers - In each layer we put the required functions only
1.2 The functions that deal with files/directory and exist in ring_state.c exist there because they are used by internal Ring Implementation (Like for finding/loading libraries) - not everything should be there.

(2) Move ring_direxists(), ring_getpathtype() to ring_vmfile.c 

(3) With respect to the new ring_fexists() implementation - put it under another name (inside ring_vmfile.c) like ring_fexists_general() then update fexists() to call ring_fexists_general()

idrassi:
Thank you for the explanations about the architecture, I see your point.
I modified the code to follow your suggestions. I think it is cleaner now.

MahmoudFayed:

Could you move ring_fexists_general() from ring_state.c to ring_vmfile.c
This pull request should update (ring_vmfile.h & ring_vmfile.c) only without any updates to (ring_state.h & ring_state.c)

Modules like (File Functions) are optional
A customized build from the source code could exclude it
By setting #define RING_VM_FILE 0
In https://github.com/ring-lang/ring/blob/master/language/include/ring_ext.h

In this case (ring_fexists_general()) will become (DEAD CODE, i.e. not used) if it still exist in ring_state.c

idrassi:
Thank you for the explanation, I learned a new thing!
I moved the function as requested.

MahmoudFayed:
Thank you very much for these nice functions
Keep up the GREAT WORK

PR #931, opened 2020-12-30
Update Socket : using set null pointer

abdallah-elsharif:


MahmoudFayed:
Thanks for the update

PR #933, opened 2021-01-03
factor 2x to 3x speedup of string handling by avoiding unnecessary allocations, reusing variables in VM stack and other optimizations

idrassi:
Existing code was allocating temporary strings in places were existing stack variables could be used.
Also, in functions returning a string, unnecessary allocation of temporary string was done instead of directly allocating the return stack variable.

Also string find/dup functions were optimized to use memcmp/memcpy instead loops. And the threshold of using memcpy was reduce from 64 to 16 based based on glibc 2.8 implementation: https://github.com/lattera/glibc/blob/master/string/memcmp.c#L53

As a benchmark on a test machine, gencode.bat of ringQt takes 14 seconds using the new code instead of 35 seconds using the previous code (2.5x speedup)


MahmoudFayed:
Thank you very much for this pull request, I looked at the updates, looks very nice, Also I will do more tests to check it 

For now, could you please update it according to the next suggestions

(1) Change ring_itemarray_getstring_raw
To ring_itemarray_getstringraw
After the project (ring) and the module like (itemarray) or modules like (VM_STACK) we don't use any more underscores

(2) Change RING_API_STACK_STRINGRAW
To RING_API_GETSTRINGRAW
In Ring API we don't use the word (STACK) it's related to implementation details of the Ring VM

(3) Change RING_VM_STACK_STRINGRAW 
To RING_VM_STACK_GETSTRINGRAW 
Adding the verb (GET) looks like other instructions (PUSH), (POP), etc.

(4) The next code is repeated many times and includes (16) as a magic number

	if ( nSize < 16 ) {
		for ( x = 0 ; x < nSize ; x++ ) {
			cString[x] = cStr[x] ;
		}
	}
	else {
		memcpy (cString, cStr, nSize);
	}

Could you define it as a preprocessor macro that work like a function called RING_MEMCPY() 
And add this macro to string.h file
Then reuse it in one line of code like this 

	RING_MEMCPY(cString, cStr, nSize);

And as a convention it will use the (int x) variable during the for loop

idrassi:
Thank you for the feedback.
I have pushed a commit with your suggestions.

MahmoudFayed:
Thanks for the update

(1) Stability Analysis : I checked all of the tests & did manual tests too ---> PASS without any problems which is very nice

(2) Performance Analysis

2.1 Using the current Ring 1.4 (under development) and before merging this pull request, I executed CodeGen for RingQt for 20 times - I get the next results in seconds (18.35, 18.53, 22.98, 18.36, 18.64, 22.29, 19.00, 18.46, 19.02, 18.37, 18.48, 22.08, 18.27, 18.05, 18.46, 19.13, 18.26, 18.81, 18.43, 18.00)
Max time	: 22.98 seconds
Min	time : 18.00 seconds
Average = 381.97 / 20 = 19.10 seconds

2.2 After merging this pull request, the results of executing CodeGen for RingQt for 20 times in seconds 
are (21.52, 21.68, 17.23, 21.77, 21.11, 17.47, 21.00, 21.92, 17.00, 17.15, 15.34, 16.90, 17.87, 16.87, 17.08, 17.17, 16.83, 17.18, 16.94, 17.05)
Min time = 15.34 seconds
Max time = 21.92 seconds 
Average = 367.08 / 20 = 18.35 seconds

2.3 The gain in performance in average = 4% - It's not too much but still a good performance gain and we already are comparing what you did in this pull request with what we have based on your previous updates which increased the performance for CodeGen by 4x (i.e. 400%) - So now it's 404% compared to Ring 1.13 

(3) Code Size Analysis

3.1 Before merging this pull request
src folder :  18820 LOC  -  include folder :  1826 LOC - Total = 20646 Loc

3.2 After merging this pull request
src folder : 18766  (Less by 54 lines)
include 1842 (increased by 16 lines)
Total = 20608

3.3 Total decrease in lines of code by 38 LOC
Percentage = (38/20646) * 100 =  0.1% ---> Which is another good thing too.

(4) Code Quality Analysis

The Usage of RING_VM_STACK_GETSTRINGRAW  is very nice idea and this is the most thing that I like about this pull request
I consider it an update towards better Code Quality with good effects on the code size and the performance too.

Keep up the GREAT WORK

idrassi:
Thank you for the detailed feedback. Actually, the performance numbers I gave were wrong because I forgot that I enabled `-O2` switch when compiling the new binary after modification while the original code was using standard build script without `-O2`.

I think ring build scripts should use `-O2` switch since it gives a factor 2x speedup for free.

MahmoudFayed:
You are welcome,  Yes, I am using ( /02 on Windows).
Fell free to update the building scripts on Linux at any time and send a pull request and I will accept it

PR #934, opened 2021-01-03
Fix crash on Linux in zip_getfilenamebyindex of RingZip caused by returning address of local variable

idrassi:
The function `zip_getfilenamebyindex` was returning the field `m_filename` of the local variable `info` but since `info` is on the stack, its content is not valid after the function returns.
The new implementation returns a copy if the field using `strdup` and then it frees this pointer at the end of `ring_zip_getfilenamebyindex`.

The code below demonstrate the crash under Linux 64-bit:

```
load "ziplib.ring"

name = exefilename()
z = new zip
z.SetFilename("test.zip")
z.open("w")
z.addfile(name)
z.close()

z = new zip
z.SetFilename("test.zip")
z.open("r")
? "count = " + z.FilesCount()
? "file = " + z.GetFileNameByIndex(1) /* crash line */
z.close()
```


MahmoudFayed:
Thanks for the update
Keep up the GREAT WORK

PR #936, opened 2021-01-03
Fix operator '=' giving wrong result for strings containing binary data

idrassi:
The code used `strcmp` to compare strings but if the binary data contains a zero then `strcmp` will have wrong logic.
For example, the program below will wrongly display that `c1` and `c2` are equal in most cases although they are different:

```
load "openssllib.ring"

failures = 0

for i = 1 to 256 {
	c1 = randbytes(256)
	c2 = c1 + randbytes(256)

	if c1 = c2 {
		failures++
	}
}

if failures != 0 {
	? "Error: " + failures + " Tests Failed!!"
else
	? "Test OK"
}

```
A typical output is:

`Error: 158 Tests Failed!!`

This change makes the code use `memcpy` preceded by a check on the string size


MahmoudFayed:
Thanks for the update
Keep up the GREAT WORK

PR #937, opened 2021-01-04
Use optimized build on Linux and macOS by using -O2 switch for gcc/clang

idrassi:
We also modify function `GenerateBatchGeneral` ring2exe.ring to add `-O2` switch to generated build scripts.


MahmoudFayed:
Very nice
Thanks for the update
Keep up the GREAT WORK

PR #939, opened 2021-01-08
Fix bug in isWindows64 and add function getarch to return architecture of the Ring executable

idrassi:
If ring.exe is 64-bit, then obviously Windows is 64-bit but the code for `isWindows64` was calling `IsWow64Process` which returns false if the process is 64-bit since in this case Wow64 is not enabled for this process (it is native 64-bit).

The fix of this bug uses the define `_WIN64` which exists when building in 64-bit mode (it is defined by both visual studio c++ and mingw64-g++).

We also add a new function `getarch() `which return the build architecture of the Ring executable.
The implementation of this function uses standard define values from across all known compilers to detect the build architecture of the Ring executable.

### GetArch() Function

We can detect the architecture of the Ring executable using the `GetArch()` function

**Syntax:**

GetArch() ---> String containing the name of the architecture of the Ring executable. Possible values are:
   - x86
   - x64
   - arm64
   - arm
   - unknown

**Example:**

```
switch getarch()
case "x86"
        ? "x86 32bit architecture"
case "x64"
        ? "x64 64bit architecture"
case "arm64"
        ? "ARM64 64bit architecture"
case "arm"
        ? "ARM 32bit architecture"
else
        ? "Unknown architecture"
end
```

### DIFF
--- a/language/include/ring_vmos.h
+++ b/language/include/ring_vmos.h
@@ -30,6 +30,8 @@
 	void ring_vm_os_chdir ( void *pPointer ) ;
 
 	void ring_vm_os_exefolder ( void *pPointer ) ;
+
+	void ring_vm_os_getarch ( void *pPointer ) ;
 	/* Constants */
 	#define RING_VM_POINTER_LIBNAME "os"
 	#ifdef _WIN32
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Thank you very much for the update and for writing the getarch() function

PR #942, opened 2021-01-12
Update SystemCmd()

abdallah-elsharif:
remove result file (cmd.txt) after get command value
remove nl from last line of the value for easily use the value 
for example
currentDir = SystemCmd("pwd")
? dir(currentDir)

Before the update we would get the value like this "/root/user/\n"
which would causes an error if we use dir() function or any other function

### DIFF
--- a/language/src/ring_file_extension.c
+++ b/language/src/ring_file_extension.c
@@ -642,6 +642,7 @@ void ring_vm_file_dir ( void *pPointer )
 						}
 					}
 				} while (FindNextFile(hFind, &fdFile))  ;
+				FindClose(hFind);
 				RING_API_RETLIST(pList);
 			}
 			else {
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Hello Abdallah

You don't need to use the Str2List() function and process the list item by item to do this update
Just use the string function
Like the Right() function : https://ring-lang.github.io/doc1.13/strings.html#right-function

MahmoudFayed:
Example
  cName = "Ring Language*"
  ? cName
  if right(cName,1) = "*"
    cName = left(cName, len(cName)-1 )
  ok
  ? cName   # print Ring Language

Online Version of this sample : https://tio.run/##K8rMS///P9kvMTdVwVZBKQjIVfBJzEsvTUxP1VLislcAS3FlpikUZaZnlGiAuTqGmiDFQHkFBZjWnNQ0mCyQnQdhauoaKmhy5WfDzFFQUFBWKCjKzCtRQLHp/38A

PR #944, opened 2021-01-14
Better Visual Studio build scripts that allow selecting target platform and debug mode, and support VS 2019

idrassi:
First enhancement applies to locatevc.bat which now accepts parameters in the format : 
`locatevc.bat [x86|x64|arm|arm64|auto] [debug]`

If locatevc.bat is invoked without any parameter, it will default to standard x86 32bit build as before.
The "auto" parameter instruct locatevc.bat to detect the architecture of the existing ring.exe and use it as a target.
The file `buildarch.ring` is used by locatevc.bat to detect the architecture of the existing ring.exe binary.

The Ring build scripts (buildvc.bat, buildvcw.bat, buildvcstatic.bat, buildtest.bat) take the same parameters as locatevc.bat.
They were enhanced to use compilation flags automatically set by locatevc.bat depending on the arguments passed.

All scripts now use `setlocal`/`endlocal` in order to avoid polluting the environment of the console used for the build

Similar enhancements were implemented for build script of ring2exe and ringconsolecolors extension. Other extensions can be enhanced the same way to allow building for x64 or arm64 but for now they will just build as x86 32bit as before.


MahmoudFayed:
Very nice update
Keep up the GREAT WORK

MahmoudFayed:
Note: It's better to move buildarch.ring from ring/language/src folder to ring/bin folder
The idea is to keep ring/language/src folder without *.ring files  + buildarch.ring needs ring.exe and having them in the same folder will be more nice.

idrassi:
OK, I will move buildarch.ring.

I have some remarks concerning the library dependencies in "extensions\libdepwin":

- Not all of them have 64-bit version. I'm building my own for missing ones (e.g. openssl and libcurl)
- There is a mix of VC runtime dll between various libraries. I noticed that official ones (e.g. mysql, allegro, freglut, sdl..) are linked against standard msvcrt.dll but others are linked against various version and some of them are even linked in debug mode. Probably we need to harmonize runtime used by building using same toolset. A good way is to build using mingw when possible since it links against msvcrt.dll or build the dlls with static runtime when using visual studio.

MahmoudFayed:
Very nice suggestions
Thank you very much :D

PR #945, opened 2021-01-14
Move buildarch.ring to "bin" folder for better organization

idrassi:


MahmoudFayed:
Thanks for the update

PR #946, opened 2021-01-15
Update Socket

abdallah-elsharif:


MahmoudFayed:
Thanks for the update
Just change this condition

if(!RING_API_ISCPOINTER(1) && !RING_API_ISNUMBER(2) && !RING_API_ISNUMBER(3) && !RING_API_ISNUMBER(4))

To

if(! ( RING_API_ISCPOINTER(1) && RING_API_ISNUMBER(2) && RING_API_ISNUMBER(3) && RING_API_ISNUMBER(4) ) )

Because the first one will display the error message only if all parameters are wrong
But the second one will display the error message if any of the parameters is wrong

abdallah-elsharif:
Hello Mahmoud
i fixed the problem thank you for the advice

MahmoudFayed:
Thanks, Keep up the GREAT WORK :D

PR #947, opened 2021-01-15
Ring2E to Ring

Ilir-Liburn:
Hello,

I'm making pull request to get Ring2E and the Ring even when it comes to API because Ring2E should be all about performance. If such request is accepted, be aware that I tested this only on Windows, Linux and Mac builds.

Changes:

Updated: random function by MS cl uses only 15 bits, now it is using 31 bit to match gcc/clang
New: randomize function which is generating 64 bit random number (without srandom)
New: nofprocessors function which returns number of cores (physical/logical)
New: uptime function which returns how long system has been running using high performance timer (0.1 ms resolution)

Uptime is also available as a general function, it is used by the Ring2E to measure performance, etc. And here is a .ring file to test all of this

```ring
load "stdlib.ring"
load "typehints.ring"

CheckOverflow(true)
NUM_DECIMALS = 9

decimals(NUM_DECIMALS)

? "No. processors = " + nofprocessors()

? "Uptime = " + uptime()

? "Randomf % Base = " + randomf(NUM_DECIMALS)

? "Randomize (64bit) = " + randomize()

double func randomf(dec)
	double time = EpochTime(Date(), Time())
	? "Epoch time = " + time
	double base = pow(10, dec)
	? "Base = " + base
	srandom(time)
	double rnd = random(base)
	? "Random (32bit) = " + rnd
	return rnd/base
```

### DIFF
--- a/language/include/ring_general.h
+++ b/language/include/ring_general.h
@@ -10,6 +10,10 @@
 #else
 	#include <unistd.h>
 	#define GetCurrentDir getcwd
+	#if __MACH__
+		/* Mac OS X */
+		#include <mach-o/dyld.h>
+	#endif
 #endif
 /* General Functions */
 
@@ -36,10 +40,3 @@ int ring_general_isobjectfile ( const char *cStr ) ;
 RING_API void ring_general_printline ( void ) ;
 
 void ring_general_showtime ( void ) ;
-
-double ring_general_uptime ( void ) ;
-/* Constants */
-
-#define	MILISEC 1000
-#define	MICROSEC (MILISEC*MILISEC)
-#define NANOSEC (MICROSEC*MILISEC)
--- a/language/include/ring_general.h
+++ b/language/include/ring_general.h
@@ -10,10 +10,6 @@
 #else
 	#include <unistd.h>
 	#define GetCurrentDir getcwd
-	#if __MACH__
-		/* Mac OS X */
-		#include <mach-o/dyld.h>
-	#endif
 #endif
 /* General Functions */
 
@@ -40,3 +36,10 @@ int ring_general_isobjectfile ( const char *cStr ) ;
 RING_API void ring_general_printline ( void ) ;
 
 void ring_general_showtime ( void ) ;
+
+double ring_general_uptime ( void ) ;
+/* Constants */
+
+#define	MILISEC 1000
+#define	MICROSEC (MILISEC*MILISEC)
+#define NANOSEC (MICROSEC*MILISEC)
--- a/language/include/ring_generallib_extension.h
+++ b/language/include/ring_generallib_extension.h
@@ -191,4 +191,7 @@
 	void ring_vm_generallib_addsublistsbymove ( void *pPointer ) ;
 
 	void ring_vm_generallib_addsublistsbyfastcopy ( void *pPointer ) ;
+	/* Thread safe */
+
+	void ring_vm_generallib_randomize ( void *pPointer ) ;
 #endif
--- a/language/include/ring_generallib_extension.h
+++ b/language/include/ring_generallib_extension.h
@@ -194,4 +194,6 @@
 	/* Thread safe */
 
 	void ring_vm_generallib_randomize ( void *pPointer ) ;
+
+	void ring_vm_generallib_uptime ( void *pPointer ) ;
 #endif
--- a/language/include/ring_os_extension.h
+++ b/language/include/ring_os_extension.h
@@ -43,11 +43,29 @@
 	void ring_vm_os_sysset ( void *pPointer ) ;
 
 	void ring_vm_os_sysunset ( void *pPointer ) ;
+
+	void ring_vm_os_uptime ( void *pPointer ) ;
+
+	void ring_vm_os_nofprocessors( void *pPointer ) ;
+	
 	/* Constants */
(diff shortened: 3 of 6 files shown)


MahmoudFayed:
Very nice functions,  Thank you very much

Just don't update ring_general.h and ring_general.c, They contains only functions used by Ring Compiler implementation where the extensions are optional (plugins) and putting a function in ring_general.c that are used only by extensions will lead to DEAD CODE if these extensions are removed.

So just move the updates to ring_general.h an ring_general.c to the modified extensions 
ring_generallib_extension.c and ring_generallib_extension.h 
ring_os_extension.c and ring_os_extension.h

Ilir-Liburn:
Ok, will do. I'm not sure if I should make uptime Ring API version returning seconds as a base and 0.1 ms resolution as precision part of the double. What do you think?

MahmoudFayed:
Since we have the Clock() function, It's better to write uptime() in Ring code using (Clock() and Clockspersecond() functions)

Ilir-Liburn:
Clock-Clokspersecond functions are not giving time since last restart end they are using millisecond resolution, while new functions are using 100 nanoseconds (0.1 microsecond) resolution. That's (1000 * 1000) / 100 precise. Will leave it as it is.

MahmoudFayed:
Very nice, Keep up the GREAT WORK

Ilir-Liburn:
I finished updating code, only os and generallib extensions are affected now.

MahmoudFayed:
Thanks for the update

MahmoudFayed:
Note: In this code in the GeneralLib extension
void ring_vm_generallib_uptime ( void *pPointer )
{
	RING_API_RETNUMBER(ring_vm_os_uptime());
}
It's just a wrapper around ring_vm_os_uptime() in OS Extension
This means dependency between GeneralLib extension and OS extension which is the right thing to do according to the architecture.
So, It's better to move everything related to uptime() from GeneralLib extension to OS Extension

Ilir-Liburn:
The point is: I use ring_something_uptime to measure instruction performance (-ins argument), so I need it under VM level, not user API. So I moved uptime needed by the VM to another source file because of the name clash. BTW, shouldn't "Print instruction operation code before execution" be "... during execution"?

I think I need to make another update. The problem is "#if defined __MACH__ && __MAC_OS_X_VERSION_MIN_REQUIRED < 101200" line. It is not a problem if we use Mac OS under 10.12 (I use Yosemite) to compile Ring, but if Ring is built using newer Mac OS, executable is going to fail on older systems because it is linked to OS defined clock_gettime.

We need something like ring_clock_gettime function to be always used on the Mac, while for the other systems, we can use something like #define ring_clock_gettime clock_gettime.

MahmoudFayed:
Hello
>> The point is: I use ring_something_uptime to measure instruction performance (-ins argument), so I need it under VM level
Special (Development Needs) is different from (User Needs) - what we add to the extensions (At least the basic extensions) must be very important and very general to be part of the basic knowledge about the language
** Less is More **

(1) Noprocessor() function is very good - I documented it too
(2) Randomize() & Uptime() - Not documented so we can think more about them - For now Randomize() doesn't work on Android

>> "BTW, shouldn't Print instruction operation code before execution" be "... during execution"?
If you mean the sentence meaning in (English Language), We have the choice 
(1) During *Program* execution
(2) Before *Instruction* execution

>> "but if Ring is built using newer Mac OS, executable is going to fail under older systems because it is linked to OS defined clock_gettime"

No problem, because we have full control on these things, we develop Ring and we provide the official binary release

Ilir-Liburn:
Isn't Android based on the modified version of the Linux kernel? What is not there: clock_gettime or rand_r or both?

BTW, I think there is a better name for a calc_uptime - get_uptime.

MahmoudFayed:
>> "Isn't Android based on the modified version of the Linux kernel? What is not there: clock_gettime or rand_r or both?"

The error message was about rand_r() function 

>> "BTW, I think there is a better name for a calc_uptime - get_uptime."

Fell free to change it, but follow this naming convention which is used in all of the ring source code
ring_modulename_submodulename_functionname
i.e. no more underscores after the sub module ( getuptime instead of get_uptime )

Ilir-Liburn:
According to this

https://www.androidos.net.cn/android/10.0.0_r6/raw/external/compiler-rt/test/msan/rand_r.cc

rand_r exists since 10.0.0_r6

MahmoudFayed:
RingLibSDL for Android uses android-ndk-r10c
The update is on my to-do list, maybe for Ring 1.15

Ilir-Liburn:
I found this advice on the internet, maybe it is relevant:

For those having trouble with srand/atof on NDK r10c.

Make sure you set your android target to 19. Your app should then work on Android 5 (21) as well as all lower versions.

If you set your android target to 21, then your app will ONLY run on Android 5. Your app will however no longer work on all lower android versions, because a lot of stdlib functions cannot be found (like srand/atof).

Ilir-Liburn:
BTW, it is possible to get Random number on Android using /dev/urandom

fd = open(/dev/urandom, O_RDONLY))

Alternative is to use TLS random generator or maybe to call Java API, but both might be related to C++ rather than C language.

MahmoudFayed:
Thanks, I will revise these points

PR #948, opened 2021-01-16
Fix various crashes in RingAllegro caused by wrong generated code and wrong function type

idrassi:
The code generator `parsec.ring` generates wrong code for function arguments in the form "TYPE var[]": it ignores the "[]" and generates the code "`TYPE var`" instead of "`TYPE* var`". This leads to crashes.
The fix is to replace all declaration of the form "`TYPE var[]`" in allegro.cf to "`TYPE *var`".

Also,  `al_run_detached_thread` was called with the wrong type of function pointer and this also was causing a crash.
The following program demonstrates the crash:

```
Load "gamelib.ring"

al_run_detached_thread("Thread1()")
al_rest(2)

Func Thread1
        for x = 1 to 5
                see x + nl
        next
        See "Thread1 finished"
```


MahmoudFayed:
Thank you very much

PR #955, opened 2021-01-23
Fix crash in ringCurl caused by usage of [] in declaration of curl_formadd@7

idrassi:
Usage of `[]` in function declaration generates wrong C code.
To fix, we replace it with `*` as a simple pointer which is equivalent


MahmoudFayed:
Thanks for the update

PR #956, opened 2021-01-24
PUSHPLOCAL

Ilir-Liburn:
Hello,

I made my version of the PUSHPLOCAL, e.g. how I think it should work. However, -icfinal is not showing updated instruction (LOADA->PUSHPLOCAL) and wawingcubes(faster) is crashing after -icfinal. Can you check this out?

To simply say: nothing is working as I expected.

### DIFF
--- a/language/src/ring_vm.c
+++ b/language/src/ring_vm.c
@@ -200,7 +200,7 @@ VM * ring_vm_new ( RingState *pRingState )
 	/* File name in the class region */
 	pVM->cFileNameInClassRegion = NULL ;
 	/* Control Performance Instructions */
-	pVM->lUsePushPLocal = 0 ;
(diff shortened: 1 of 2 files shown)


Ilir-Liburn:
OK, wavingcubes(faster) is crashing if I terminate it with ALT+F4, otherwise doesn't (mouse click)
I made version of wavingcubes without thread, just calling function via eval. Now I can see that LoadA is replaced by PushPLocal, altough not everywhere I wanted. From this I can assume that it is replaced in a threaded version as well, since thread runs through eval, but -icfinal is not printing replaced instructions? Or it is something other going on there because I don't see performance change at all.

MahmoudFayed:
Thank you very much, I will check it

Ilir-Liburn:
Closing due to lack of feedback.

PR #959, opened 2021-03-28
fix for infinite recursions

Ilir-Liburn:
Hello,

here is a pull request as you suggested. Please note, I also changed showerrormessage function, otherwise long list of (empty in content) messages are printed to the console. This can also be fixed by simply putting

`		if ( x == RING_VM_STACK_SIZE ) {
			break ;
		}
`

at the end of the loop. Which one is correct, you decide.

Back to newscope function: I'm not sure, but maybe we should use RING_VM_STACKSIZE multiplied by 4 here. If I remember correctly, stack size is usually 4096 bytes on 32 bit systems, address takes 4 bytes, therefore 1024 (return) addresses can fit onto stack (thinking about Ring2C here). However, this has to be documented then. If someone increases RING_VM_STACKSIZE to 1024, then above does not apply I think.

### DIFF
--- a/language/src/ring_vm.c
+++ b/language/src/ring_vm.c
@@ -1121,6 +1121,7 @@ void ring_vm_callclassinit ( VM *pVM )
 RING_API void ring_vm_showerrormessage ( VM *pVM,const char *cStr )
 {
 	int x,lFunctionCall  ;
+	char *cStr2 ;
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Thanks for the update
Keep up the GREAT WORK

PR #961, opened 2021-07-04
deployment issue

alas82:
hi , thanks for this good language and wish you more progress in your road to success 
i tried to distribute small program , it copied too many dependency DLL which make size too big , is there anyway to make more small applications?
also if ring include qtcharts will be good add on to it.

thanks


MahmoudFayed:
Not a pull request
Please submit your questions to Ring Group

PR #965, opened 2021-07-16
to be on the safe side

Ilir-Liburn:
Hello Mahmoud,

it seems there is some fix from Ring2A I could bring here. Fixing application crash when main thread and worker thread are in conflict. If you remember, the colors list initialization in wavingcubes sample cause app crash if it is done after the creation of the thread. It seems that main thread and worker thread are trying to write into GC pool memory at the same time. To be on the safe side, I fixed it by completely removing memory pool usage at ringlibuv. Please review, see if can be done better (differently).

There is also conflict in a worker thread code. At the end of the execution, there is a code which is bringing back items to the memory pool of the main thread causing write operation on both sides. This is fixed in Ring2A, but I can't bring it here because my code is significantly different.


MahmoudFayed:
Thanks for the update

PR #966, opened 2021-08-19
better ringCMD.bat 

matan-h:
Hi,
I fixed the `RingCmd.bat` to work without `cd` to the script folder (it use `%cd%`) and also display the ring help message.
Thanks, 
Matan h


MahmoudFayed:
Thank you very much

PR #968, opened 2021-09-11
hotfix2

Ilir-Liburn:
Hello,

I fixed tmpnam_s problem, functions was trying to return absolute path, but there wasn't enough space in buffer.


MahmoudFayed:
Thanks for the update

PR #969, opened 2021-09-16
hotfix3

Ilir-Liburn:
Hello,

STACK_POP moved before return


MahmoudFayed:
Thanks for the update

PR #971, opened 2021-09-30
fixed installer and uninstaller

aurora:
Hello,
as written in the group i fixed some issues with installer and uninstaller:

* ln now uses -f flag to force creation of link (overwrite link if it already exists)
* added xattr -d com.apple.quarantine to remove quarantine flag on macOS
* updated uninstaller to remove symlinks accordingly on macOS

btw.: contrary to what i wrote in the mailing-list, all links to /usr/local are correct in the installer for macOS (i got a little confused with the linux/macOS installation pathes)

### DIFF
--- a/bin/install.sh
+++ b/bin/install.sh
@@ -5,459 +5,459 @@
 
 if [ -f ../lib/libring.dylib ];
 then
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Thanks for the update

PR #988, opened 2021-10-15
fixed zero case

Ilir-Liburn:
Hello Mahmoud,

here is the "zero case" fix. Please note: greater-equal, less-equal, etc. is not covered, check if it is necessary. Issue is complex and it should be revised for Ring 1.16 I think.


MahmoudFayed:
Hello Ilir

Thanks for the update
I will revise/test the different cases and will do any required updates

Keep up the GREAT WORK

PR #1079, opened 2021-11-07
Fixed Unknown property backgound-color

pancakevirus:


MahmoudFayed:
Thanks for the update

PR #1086, opened 2021-11-07
Update factorial_tail_recursionController.ring

CalmoSoft:
Please accept the change.
(See in Ring Group)


MahmoudFayed:
Thanks for the update
Note: Using (Return) is not necessary

PR #1106, opened 2021-11-10
fixed typo, mistranslation...

pancakevirus:
Also same error in v2, v3 - AA-Sudoku.ring

...and other typo:
 * Inputt → Input
 * Outputt → Output


MahmoudFayed:
Thanks for the update

PR #1137, opened 2021-11-11
prepaired next version log.

pancakevirus:
Currently version number is not decided. (Ring2A/2B based, v.2.x?)


MahmoudFayed:
It will be Ring 1.17 - The release date still needs to be determined

PR #1216, opened 2021-11-14
Update binary_heap.ring

pancakevirus:
fixed typo.


MahmoudFayed:
Thanks for the update

PR #1225, opened 2021-11-14
Update mathspeed3.ring

pancakevirus:
fixed typo.


MahmoudFayed:
Thanks for the update

PR #1229, opened 2021-11-14
fixed typo.

pancakevirus:
Also, https://github.com/ring-lang/ring/search?q=deleteing


MahmoudFayed:
Thanks for the update

PR #1275, opened 2021-11-19
fixed.

pancakevirus:
- Removed non-ASCII odd character. (  ِِApplication )
 - Reformatted.
 - Remove illegal whitespace from filename. (" test05 short way for sum .ring" → "test05 short way for sum.ring") - make sure " test06 put&get .ring" in Windows Explorer.


pancakevirus:
same_problem

MahmoudFayed:
Thanks for the update

PR #1349, opened 2021-11-22
Update and rename Leson 2.ring to Lesson 2.ring

pancakevirus:


MahmoudFayed:
Thanks for the update

PR #1365, opened 2021-11-23
Update and rename start to getting_started.ring

pancakevirus:


MahmoudFayed:
Thanks for the update

PR #1392, opened 2021-11-23
Update oop_version_1.ring

pancakevirus:


MahmoudFayed:
Thanks for the update

PR #1416, opened 2021-11-24
Create CODE_OF_CONDUCT.md

pancakevirus:
To make it easier to contribution.

If you don't need it, when reject it.


MahmoudFayed:
I agree with you, It's a nice addition
Thank you very much

PR #1421, opened 2021-11-30
Update Sockets: extend constants from c instead of define it hard coded

abdallah-elsharif:
Socket constants are different in operating systems, so extending them is better than defining them hard coded to avoid wrong values in different systems .


MahmoudFayed:
Thanks for the update

PR #1422, opened 2022-01-03
safe pointer

Ilir-Liburn:
Hello,

here is the commit, needs testing


MahmoudFayed:
Thanks for the update

PR #1424, opened 2022-01-16
list decimal print fix

Ilir-Liburn:
Hello,

here is the fix, greetings


MahmoudFayed:
Thanks for the update

PR #1428, opened 2022-03-19
Ring Scanner: avoid unnecessary strdup call in ring_scanner_checktoken

idrassi:
This gives a performance boost of 4% on Windows for large ring files

### DIFF
--- a/language/src/ring_scanner.c
+++ b/language/src/ring_scanner.c
@@ -443,13 +443,19 @@ void ring_scanner_checktoken ( Scanner *pScanner )
     /* This function determine if the TOKEN is a Keyword or Identifier or Number */
     assert(pScanner != NULL);
     /* Not Case Sensitive */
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Thanks for the update

MahmoudFayed:
I did a little update in this commit : https://github.com/ring-lang/ring/commit/73dbc87ea18375403998cf3ca62b661794c93efa

Since we already test the condition ( pScanner->pRingState->lNotCaseSensitive ) and we have (else) - No need to test (! pScanner->pRingState->lNotCaseSensitive) using another if statement.
Repeating the function call ( ring_hashtable_findnumber() ) is better than repeating the if statement.

idrassi:
Thank you for the update.
Indeed, this is better!

MahmoudFayed:
You are welcome

PR #1429, opened 2022-03-19
Add function Getfilesize that is more reliable & faster than Fsize from Stdlib.ring

idrassi:
Benchmarks show that this new function is at least 20% faster than `Fsize` that is available in `Stdlib.ring`.
Its implementation uses the C `stat` function that gets the information about file size directly from the filesystem without requiring any I/O. This makes it much faster than using `fopen->fseek->ftell->fclose`.
Moreover, there are situations where `fopen` will fail because of sharing violation and so `fseek`/`ftell` cannot be used to get the file size in this case. The C `stat` function doesn't suffer from this issue.

### **Getfilesize() Function**

We can get the size in bytes of a given file using the Getfilesize() function

**Syntax:**

Getfilesize(cFilePath) ---> file size in bytes as a positive Number or -1 in case of failure (e.g. path doesn't exist or not a regular file)

**Example:**

```
? Getfilesize("b:\ring\bin\ring.exe") + nl +
	  Getfilesize("b:\ring") + nl +
	  Getfilesize("b:\ring\ring2.exe") 
```

**Output:**

```
80384
-1
-1
```

### DIFF
--- a/language/include/ring_file_extension.h
+++ b/language/include/ring_file_extension.h
@@ -64,6 +64,8 @@
     void ring_vm_file_direxists ( void *pPointer ) ;
 
     void ring_vm_file_getpathtype ( void *pPointer ) ;
+	
+	void ring_vm_file_getfilesize ( void *pPointer ) ;
     /* Number & Bytes */
 
     void ring_vm_file_int2bytes ( void *pPointer ) ;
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Nice addition
Thank you very much

PR #1431, opened 2022-03-26
RingLibUI: Fix memory leaks and memory corruption related to wrong usage of uiFreeText

idrassi:
The sample test3.ring was crashing on Linux because of call to `uiFreeText`. This function should not be called from Ring code since it is an internal libUI function responsible for freeing memory returned by various functions (like `uiOpenFile`) and it should not be called on Ring string pointers.

Moreover, many libUI functions returning "`char *`" were leaking memory because `uiFreeText` was not called after copying the string to Ring. For these functions, like `uiWindowTitle`, dedicated implementation is added libui.cf that calls `uiFreeText` after copying the returned string in Ring return value.

This change addresses the memory leaks and removes `uiFreeText` from the list of libUI function visible to Ring code.

### DIFF
--- a/documents/build/html/_sources/libui.txt
+++ b/documents/build/html/_sources/libui.txt
@@ -364,7 +364,6 @@ Control Gallery
 			return
 		ok
 		uiEntrySetText(entryOpen, filename)
-		uiFreeText(filename)
 
 	func onSaveFileClicked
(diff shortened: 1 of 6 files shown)


MahmoudFayed:
Very nice update
Thank you very much

PR #1432, opened 2022-03-31
Trycatchfix

Ilir-Liburn:
Hello,

here is the fix.

Greetings,
Ilir


MahmoudFayed:
Thanks for the update

PR #1433, opened 2022-04-02
Fix Index out of range error in AppPath function when application is compiled by Ring2EXE

idrassi:
For this fix, we add a new function `IsAppCompiled` that checks whether the application has been compiled using Ring2EXE


MahmoudFayed:
Thanks for the update

PR #1434, opened 2022-04-02
Add missing winapi.ring and wincreg.ring from bin/load

idrassi:
Without these files, the statements '`Load "winapi.ring"`' and '`Load "wincreg.ring"`' fail despite the fact that ring_winapi.dll and ring_wincreg.dll are part of official Windows distribution.


MahmoudFayed:
Thanks for adding the files

PR #1435, opened 2022-04-02
Fix uptime bug on Windows if high resolution timer frequency is more than 10Mhz

idrassi:
`QueryPerformanceCounter` returns the number of ticks of the high resolution timer and the previous code was assuming that its frequency is always 10Mhz which is not always true. `QueryPerformanceFrequency` must be called to get the effective frequency.

https://docs.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancecounter
https://docs.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancefrequency
https://docs.microsoft.com/fr-fr/windows/win32/sysinfo/acquiring-high-resolution-time-stamps

### DIFF
--- a/language/src/ring_os_extension.c
+++ b/language/src/ring_os_extension.c
@@ -297,9 +297,11 @@ void ring_vm_os_uptime ( void *pPointer )
 {
     double nTime  ;
     #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Thanks for the update

PR #1436, opened 2022-04-03
Windows build batch files enhancements: better setlocal usage, use 'call' in locatevc.bat to run buildarch.ring

idrassi:
The `setlocal` change avoids that environment variables are polluted which was causing failure when building a lot of extensions in the same CMD session.
The `locatevc.bat `change fixes an issue when using "auto" as first parameter because running "`RINGEXEPATH`" at line 32 was causing an error in some cases: using `call` avoids this problem and it is more reliable.


MahmoudFayed:
Thanks for the update
Keep up the GREAT WORK

PR #1437, opened 2022-04-06
Update StdLib - Add AppArguments() function

idrassi:
`AppArguments` function returns the effective arguments passed to the Ring script. This is more practical than `sysargv` when we just need the effective arguments specified by the user to the Ring script and not the exe name and script name. Moreover, depending if the Ring script is compiled or not, `sysargv` doesn't have the same format with regards to the position of the effective arguments.

### DIFF
--- a/documents/source/stdlib.txt
+++ b/documents/source/stdlib.txt
@@ -149,6 +149,33 @@ Example:
 	else see "Application is running under Ring interpreter" ok
 
 .. index:: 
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Nice function, Thanks for adding it
I remember writing GetParameters() function is RingPM source code to work around this problem
URL: https://github.com/ring-lang/ring/blob/master/tools/ringpm/core/getpara.ring

Keep up the GREAT WORK

PR #1438, opened 2022-04-09
Fix bug in handling of binary string by VM stack that caused loss of value

idrassi:
The VM stack handling of string values supposes that they are simple text string that are null terminated and this caused string values holding binary data to lose correct value under various circumstances.
For example, the following code produce incorrect result:
```
keyValue = hex2str("0011223344556677")
key = [:name="key", :value=keyValue]
See "keyValue length = " + Len(keyValue) + nl
See "key.value length = " + Len(key[:value]) + nl

if keyValue = key[:value] See "Test OK" + nl else See "Test Failed!!" + nl ok
```
Its output is:
```
keyValue length = 8
key.value length = 0
Test Failed!!
```

As we can see, we lost `keyValue` content once it was copied into the list `key`.

This commit fixes this by changing the code to use `ring_itemarray_setstring2`, `ring_string_new2_gc`, `ring_list_addstring2_gc` and  `ring_list_setstring2_gc` at places not requiring text strings in order to not loose information about the real size of the string and to avoid the use of `strlen` function on binary strings.
At functionality level, there is no change since using the real size with text strings when copying them will keep the position of the zero terminating characters and so strlen/strcmp and related functions will behave the same.

### DIFF
--- a/language/include/ring_vm.h
+++ b/language/include/ring_vm.h
@@ -132,7 +132,7 @@
     #define RING_VM_STACK_PUSHCVAR ring_itemarray_setstring2(pVM->aStack,pVM->nSP,ring_list_getstring(pVar,3),ring_list_getstringsize(pVar,3))
     #define RING_VM_STACK_PUSHNVAR ring_itemarray_setdouble(pVM->aStack,pVM->nSP,ring_list_getdouble(pVar,3))
     #define RING_VM_STACK_PUSHPVALUE(x) pVM->nSP++ ; ring_itemarray_setpointer(pVM->aStack, pVM->nSP , x )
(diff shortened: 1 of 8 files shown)


MahmoudFayed:
Thanks for the update

MahmoudFayed:
The code is revised in this commit : https://github.com/ring-lang/ring/commit/a555401bb3e5ec0200177b154b30fa841d9eab5d

RING_VM_STACK_PUSHCVALUE stays without modification
We already have RING_VM_STACK_PUSHCVALUE2() - So we use it
This allows using RING_VM_STACK_PUSHCVALUE("") instead of RING_VM_STACK_PUSHCVALUE("",0)

idrassi:
Thank you for the information. I didn't notice `RING_VM_STACK_PUSHCVALUE2`. Code is indeed cleaner this way.

MahmoudFayed:
You are welcome 
Keep up the GREAT WORK

PR #1439, opened 2022-04-16
Fix memory leak in ring_vm_loadindexaddress

idrassi:
Some call paths of `ring_vm_loadindexaddress` were existing using `return` statement without freeing the `pString` variable allocated at the beginning of the function.


MahmoudFayed:
Thanks for the update

PR #1440, opened 2022-04-16
Remove unnecessary String allocation in ring_vm_range and ring_vm_string_assignment

idrassi:
In `ring_vm_range`, we can get the size and first character of each string beforehand and so no string copy is needed.
In `ring_vm_string_assignment`, we can get the size and first character of the string beforehand and so no string copy is needed.


MahmoudFayed:
Thanks for the update

PR #1441, opened 2022-04-16
Fix bug in NotEqual operator when used with strings

idrassi:
`!=` operator returns wrong result when used with strings containing binary data.
The following Ring code reproduces the issue

```
value1 = hex2str("3ad70001")
value2 = hex2str("3ad7000198A4E841")

if value1 != value2
	See "Test OK" + nl
else
	See "Test FAILED!!" + nl
ok
```


MahmoudFayed:
Thanks for the update
Keep up the GREAT WORK

PR #1443, opened 2022-04-18
RingOpenSSL: fix openssl 1.1.1 compatibility by initializing it correctly and using adequate AES-XTS test vector that doesn't use duplicated keys

idrassi:
OpenSSL 1.1.1 rejects duplicated XTS keys, so we ensure that our test vectors follow this:
https://github.com/openssl/openssl/blob/c39352/crypto/evp/e_aes.c#L3119

Also, we need to call `OPENSSL_init_crypto` explicitly to make OpenSSL populate internal list of algorithms.


MahmoudFayed:
Thanks for the update

PR #1444, opened 2022-04-24
RingOpenSSL: first implementation of RSA support (key generation,import,export,encryption,decryption)

idrassi:
The script `test_rsa.ring` implements tests for all added functions and it can serve as a reference alongside the documentation.
Below is the list of added functions. The documentation (`secfunc.txt`) was updated with all details about these functions and examples of usage.
 - `rsa_generate`
 - `rsa_export_params`
 - `rsa_import_params`
 - `rsa_export_pem`
 - `rsa_import_pem`
 - `rsa_is_privatekey`
 - `rsa_encrypt_pkcs`
 - `rsa_decrypt_pkcs`
 - `rsa_encrypt_oaep`
 - `rsa_decrypt_oaep`
 - `openssl_versiontext`
 - `openssl_version`

One interesting point about `test_rsa.ring` is that I implemented a deep comparison function for lists (cf `Compare` in `test_rsa.ring`). It would have been helpful to have such function in the standard library.

### DIFF
--- a/documents/source/secfunc.txt
+++ b/documents/source/secfunc.txt
@@ -25,6 +25,18 @@ Before using the next functions load the openssllib.ring library
 * Encrypt()
 * Decrypt()
 * Randbytes()
+* rsa_generate
+* rsa_export_params
+* rsa_import_params
+* rsa_export_pem
+* rsa_import_pem
+* rsa_is_privatekey
+* rsa_encrypt_pkcs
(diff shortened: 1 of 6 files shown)


MahmoudFayed:
Nice contribution, Thank you very much.

>> "It would have been helpful to have such function in the standard library."
Fell free to improve the StdLib

Keep up the GREAT WORK

PR #1445, opened 2022-04-26
Stdlib: Add CheckEquality function that checks if two items are equal using deep comparison if needed

idrassi:
This function can be used to check equality of variables in a generic way and it is very useful when we need to check if two lists hold equal values.
For example, this can be used to compare the result of JSON parsing:

```
load "stdlib.ring"
load "jsonlib.ring"	

cJSON =`
{
	"name": "Ring",
	"year": 2016,
	"license": "MIT",
	"other": null
}
`

listParsed = JSON2List(cJSON)
listExpected = [:name = "Ring", :year = 2016, :license = "MIT", :other = null]

? CheckEquality(listParsed,listExpected)
```

### DIFF
--- a/documents/source/stdlib.txt
+++ b/documents/source/stdlib.txt
@@ -1521,3 +1521,30 @@ Example:
 	load "stdlib.ring"
 	aList = 1:5
 	? RandomItem(aList)
+
+.. index:: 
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Thank you very much

PR #1446, opened 2022-04-29
RingWinapi: Fix various crashes, memory leaks and other issues.

idrassi:
Some WINAPI functions were called using wrong types (e.g. `LookupAccountSidA`).
`rwaGetErrorMsg` was causing crash for long error message because of use of `sprint_s`. Now we use `StringCbPrintfA` which truncate if needed.
Error code was lost in many cases leading to wrong error displayed.
Memory leak was fixed in `ring_winapi_rwaenvirvarstring`
wrong function to free memory was used in `rwaGetErrorMsg`(`LocalFree` should be used instead of `HeapFree`)


MahmoudFayed:
Thanks for the update
Keep up the GREAT WORK

PR #1447, opened 2022-04-29
Fix runtime check failure in Debug build caused by unintialized cStr2 variable

idrassi:
This line is not needed at all since `cStr` is not meant to be changed (it is `const char*` parameter). removing this line fix the issue and doesn't change the behavior.


MahmoudFayed:
Thank you very much

PR #1448, opened 2022-04-30
RingSockets: various fixes

idrassi:
Fix memory leaks, avoid calling WSAStartup too many times and avoid closing invalid socket if socket function fails


MahmoudFayed:
Thanks for the update
Keep up the GREAT WORK

PR #1452, opened 2022-05-08
Multiple assignments

Ilir-Liburn:
Hello Mahmoud,

this is how I fixed multiple assignments problem in Ring2A/B. Feel free to review it, e.g. see if there could be issues or better way to do it.

Greetings,
Ilir

### DIFF
--- a/language/include/ring_vm.h
+++ b/language/include/ring_vm.h
@@ -153,6 +153,7 @@
     #define RING_VM_STACK_READP ring_itemarray_getpointer(pVM->aStack,pVM->nSP)
     #define RING_VM_STACK_OBJTYPE pVM->aStack[pVM->nSP].nObjectType
     #define RING_VM_STACK_PREVOBJTYPE pVM->aStack[pVM->nSP-1].nObjectType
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Hello Ilir

Thanks again for the report and the solution, I will check and do some tests

For now, When I looked at the solution I started to think about the next things

(1) Is this really a VM problem? - In other words, Can this problem solved at the Compiler side instead of the VM side to provide a more clear solution?

(2) Is defining RING_VM_STACK_PUSH the right thing to do?
It could be but I am not sure (I need to think about this)
because it will do (pVM->nSP++) which is not safe alone. because when we use pop through RING_VM_STACK_POP which will (pVM->nSP--) This still keep the real value of the item on the stack (this value could be a dangling pointer after some time).

Greetings,
Mahmoud

Ilir-Liburn:
Hello Mahmoud,

yes, it is better to do it at parsing stage, using ring_vm_dup or similar. I made this as proof of concept, it is not final.

Greetings,
Ilir

PR #1453, opened 2022-05-10
RingWinCReg: fix memory leaks

idrassi:
The "CRegistry" pointer was never freed once created.
`lpszStr` member of `CRegEntry` was not freed first before being overwritten.


MahmoudFayed:
Thanks for the update

PR #1454, opened 2022-05-20
stringregexp 

faustf:
Hi sorry  , this is my first time i use github  and  this procedure ,ofcourse  the special thanks is  for Ilir (one day I'll make you a statue) , i create  a  stringregexp , this  extention is maked in C (another special thanks at Ilir) . you can  use  a regexp in style C  in style Perl with PCRE (not PCRE2) and also  in C but  with sscanf , it's not perfect  but is perfectible , i created  because  with one  line  you can create aregexp  and  also in perl Style :) bye  and  thanks at  all


MahmoudFayed:
Hello Stefano
Thanks for your contribution
I prefer keeping this extension as an external package

faustf:
ok thanks  you too have  nice day :D

PR #1455, opened 2022-05-23
Update RingSockets: avoid calling disabled functions

abdallah-elsharif:
some functions are  disabled because they don't work on macOS but still library calls them.
https://github.com/ring-lang/ring/commit/3d2210fdf565e2fd5da44ba064fe87c8e1352904


MahmoudFayed:
Thanks for the update

PR #1456, opened 2022-05-23
Fix stack overflow crash in parser on Windows when loading large RingQt files by setting stack reserved size to 8 MiB

idrassi:
Linux uses 8 MiB reserved stack size by default whereas Windows uses 1 MiB. Setting it explicitly to 8 MiB to match Linux and to fix the parser crash.


MahmoudFayed:
Thanks for the update

PR #1457, opened 2022-05-23
Ring2Exe: Fix distributing Windows GUI applications when there is no compiler on the machine

idrassi:
Previously, `ring.exe` was always used as a base. Now we use `ringw.exe` for GUI and `ring.exe `for console.


MahmoudFayed:
Thank you very much

PR #1458, opened 2022-05-24
Add ipv6 conversion functions

abdallah-elsharif:
We have two functions that convert ipv4 (inet_addr - inet_ntoa)
but new functions can convert both ipv4 and ipv6 
thanks.


MahmoudFayed:
Thanks for the update

PR #1459, opened 2022-05-25
RingQt Windows: Fix conflict with existing Qt installations by forcing the use of Ring own distributed Qt

idrassi:
Before loading `ringqt.dll` and `ringqt_light.dll`, we set the environment variable `QT_QPA_PLATFORM_PLUGIN_PATH` to the location of our executable file since this is where all our Qt dll are located.


MahmoudFayed:
Thanks for the update

PR #1460, opened 2022-05-25
Ring Notepad: Add menu to build GUI applications with ring2exe

idrassi:
This is useful for LibUI based applications


MahmoudFayed:
If this option is useful for Libui application
Then instead of using ",-static,-gui"
It's better to use ",-dist,-gui,-libui"
So Ring2EXE can copy libui dlls

idrassi:
This menu is for building only, not distribution. We can have an additional menu dedicated to distributing LibUI application.

Today, the only build menu in RingNotepad is for console applications. This PR make it possible to compile also GUI application, thus the addition of -gui.

MahmoudFayed:
Thanks for the update

idrassi:
Thank you.
By the way, I noticed that the macOS zip file for 1.17 version cannot be opened. It seems to be corrupted. Maybe it is worth checking (uploading large files often have such issues)

MahmoudFayed:
Thanks for the report, I will check it

PR #1461, opened 2022-05-27
Update RingSockets: Add socketsCleanup function

abdallah-elsharif:
This function will terminates use of the Winsock, we have to document that to tell users call this function when they finish using the extension

Thanks


MahmoudFayed:
Thanks for the update

PR #1465, opened 2022-06-21
Added Redirect Function for Page Redirection

aaveshdev:
Redirect Function consists one parameter as a destination webpage address.


MahmoudFayed:
Thanks

PR #1466, opened 2022-07-08
fixed cursor position problem while using unicode

Ilir-Liburn:
Hello Mahmoud,

Manoj Negi reported issue reproducible by

see "Enter your name حنين:  " give me
see nl
see "You are " + me

where unicode text is causing input to be inserted before the see text. I solved this by moving the cursor to end of the line.

P.S.

I used Arabic text from the Softanza lib, don't know the meaning.


MahmoudFayed:
Hello Ilir

Thanks for the update
The word (حنين) called (Haneen)
It's an Arabic name used for girls 
Our friend Mansour have a daughter called Haneen (So the name is used in SoftanzaLib)
The name meaning (Nostalgia), i.e. Expressing love/missing for something/someone

PR #1468, opened 2022-08-12
Created Regex Extension for Ring

aaveshdev:


MahmoudFayed:
You can create the extension as a package
See this chapter: https://ring-lang.github.io/doc1.17/ringpm.html

PR #1469, opened 2022-08-12
Header File for Building Regex Extension

aaveshdev:


MahmoudFayed:
The extension should be distributed as a package using Ring package manager
See: https://ring-lang.github.io/doc1.17/ringpm.html

PR #1470, opened 2022-08-12
Query String in Post Method

aaveshdev:
Added functionality to retrieve query string in post method.


MahmoudFayed:
Thanks for the update
(1) In Ring, NULL is the same as the empty string "", so You don't need to test both of them
(2) In If Statement, Using OK is better than End to follow the same style in this source code file
(3) The variable cInput2 is not required, We can use sysget("QUERY_STRING") directly

PR #1476, opened 2022-10-07
commented out some lines in constants.ring of the ringsockets extensi…

gautxo:
…on that caused it to give an error when compiling on Mac OSX


MahmoudFayed:
Thanks for the update

PR #1477, opened 2022-10-09
fixed missing string index check

Ilir-Liburn:
Hello,

while checking this code

? ptr2str(ptr,0,4)[5]

I realized I got value outside of the string bounds, so here is the fix.

Greetings,
Ilir

### DIFF
--- a/language/src/vmlists.c
+++ b/language/src/vmlists.c
@@ -306,6 +306,17 @@ void ring_vm_listpushv ( VM *pVM )
         RING_VM_STACK_OBJTYPE = RING_OBJTYPE_LISTITEM ;
         ring_vm_oop_setbraceobj(pVM, (List *) ring_item_getlist(pItem));
     }
+    else if ( ring_item_gettype(pItem) == ITEMTYPE_POINTER ) {
+        if ( pVM->nRetItemRef > 0 ) {
+            pVM->nRetItemRef-- ;
+            if ( (ring_vm_isstackpointertoobjstate(pVM) == 1) ) {
+                RING_VM_STACK_SETPVALUE(pItem);
+                RING_VM_STACK_OBJTYPE = RING_OBJTYPE_LISTITEM ;
+                return ;
+            }
+        }
+	RING_VM_STACK_SETCVALUE2(NULL,0);
+    }
 }
 
 void ring_vm_listassignment ( VM *pVM )
--- a/language/src/vmlists.c
+++ b/language/src/vmlists.c
@@ -202,6 +202,10 @@ void ring_vm_loadindexaddress ( VM *pVM )
             }
         }
         else if ( RING_VM_STACK_ISSTRING ) {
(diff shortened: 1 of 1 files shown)


Ilir-Liburn:
Hello,

it seems git doesn't allow to pull another request on the same file, so here is the fix: while using

? ptr[1] # attempt to print pointer value
for e in ptr ? e next # attempt to iterate over the pointer list

I got VM crash because PUSHV on list item doesn't handle pointer.

Observe: I immediately decrease nRetItemRet value if > 0 because I noticed unhandled situations like

func retptr
	...
	return &ptr[1]

This is tested in Ring2B and it works. Still, I think it is not correct to return list item in case of the pointers, e..g whole if section can (should) be removed.

P.S.

I return empty string to disallow changing the pointer in situations like

ptr[1] = 0 // ptr address should be set via setPointer function, not like this

MahmoudFayed:
Hello Ilir

(1) At first this is  a complete test program that demonstrates the problem and can be used for testing

`
? "Define a number"
n1 = 12553523

? "Create a pointer to the number"
ptr = varptr("n1",:int)

? "Print the pointer value"
? ptr

? "Get a string using the pointer"
m = ptr2str(ptr,0,4)

? "String Size: " + len(m)

? "Print the ASCII code for each character"
for c in m ? ascii(c) next

? "Print character outsize range"
// ? ptr2str(ptr,0,4)[5]	# CRASH

? "Print each item in pointer list"
for e in ptr
//	? e 			# CRASH
next

? "Print the pointer value"
//	? ptr[1]		# CRASH

? "Call function that return a string"
? mystr()			# String (4 Characters)

? "Access character outside range"
? mystr()[5]			# Missing ERROR!

func mystr
	return "1234"
`

(2)  We have two problems, The first is using an index outside the string range (You already solved)
The second problem is using PUSHV with a list item that contains the pointer
The right behavior in this situation should be (Pushing a String that contains the pointer address after converting it to a string)
So using
`
for e in ptr
	? e 			
next
`
Should produce a similar result to using 
`
? ptr
`

(3) I will accept the pull request for now
After that I will change the code 

Thank you very much

Ilir-Liburn:
Hello Mahmoud,

**Thank you very much**

You're Welcome. The problem is: I don't want to have following situation

? "Define a number"
n1 = 12553523

? "Create a pointer to the number"
ptr = varptr("n1",:int)

setPointer(ptr, 3.14+ ptr[1]) // implicit conversion of a string to number 

That's why I return empty string in Ring2B, but now after I revised this, I think I will return NIL to prevent any pointer manipulation outside of the get/setPointer functions in Ring2B+, and match Ring2B to official Ring.

Greetings,
Ilir

Ilir-Liburn:
Hello Mahmoud,

I understand, maybe I'm overthinking or been overprotected ;) but my goal is to make Ring VM as a development environment taking care about such things before bringing it to compiled executable (without VM).  Somehow in my vision VM is safeguard against later runtime failures at the end of the development cycle.

PR #1478, opened 2022-10-18
Create bert-codegen.h

umariani:
Orig:  ring/language/include/codegen.h
New:  ring/language/include/bert-codegen.h   ==> codegen.h

Adds "^^" as  power function   example  4^^3 => 64
Uses  "void ring_vm_explog ( VM *pVM )"

### DIFF
--- a/language/include/bert-codegen.h
+++ b/language/include/bert-codegen.h
@@ -1,0 +1,207 @@
+/* Copyright (c) 2013-2022 Mahmoud Fayed <msfclipper@yahoo.com> */
+#ifndef ring_codegen_h
+    #define ring_codegen_h
+    /*
+    **  Data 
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Hello Bert

I don't plan to add this operator at the current stage
Will think about this in the future once I determine the approach that match the language design
For example we have many functions in this chapter : https://ring-lang.github.io/doc1.17/mathfunc.html
One of the ideas is to keep using them while introducing specific VM instruction instead of the (function call)
to get the same performance.

I will close this for now, and will check them in the future.

Thanks

MahmoudFayed:
Hello Bert

Thanks for your contribution, Now Ring 1.18 support the ^^ operator

PR #1489, opened 2022-11-11
RingNotepad - fix Running Web apps on Ubutnu and MacOSX

MajdiSobain:


MahmoudFayed:
Thanks for the update

PR #1491, opened 2022-12-07
fixed self referencing

Ilir-Liburn:
Hello  Mahmoud,

I fixed case where self referencing is not increasing reference counter. Here is a test which fails

```Ring
tmp = new Temp { data="test2" }
cls2 = [ref(tmp)]
? refcount(tmp)
? "--------"
tmp=NULL
? refcount(cls2[1])
cls2[1] { tmp = ref(self) }
? tmp
? refcount(tmp)
? cls2[1]
? refcount(cls2[1])

class Temp
	data
```

BTW, all tests are passing now in Ring2B with temp memory deletion. It was easier than I thought because of lDontDelete.
What I miss now is information about collecting cycles code added last minute. I don't have this in Ring2B, but tests are passing. How come (assuming cycle and cycle2 are corresponding tests)?

Greetings,
Ilir

### DIFF
--- a/language/src/vmgc.c
+++ b/language/src/vmgc.c
@@ -283,6 +283,11 @@ RING_API List * ring_list_newref_gc ( void *pState, List *pVariableList, List *p
         pList->gc.pContainer = pVariableList ;
     }
     else {
+        /* If we don't have references here - it is a new reference! */
+        if ( ! ring_list_isref(pList) ) {
+            pList->gc.lNewRef = 1 ;
+            ring_list_updaterefcount_gc(pState,pList,RING_LISTREF_INC);
+        }
         pVariableList = (List *) pList->gc.pContainer ;
     }
     return pVariableList ;


Ilir-Liburn:
Hello Mahmoud,

I tried samples inside infinite loop and I can't see any memory leak, although I don't use cycle detection code (because I don't understand at this time how is that possible).

But we have another problem now, following code

```Ring
tmp = []
ref(tmp) = NULL
? refcount(tmp)
```

prints 65536. This happens because reference is decreased twice if new ref is present in case of the deletion, which seems odd to me, If there was no assignment (because new ref is still present), then new ref should be  only cleared, not reference decreased once more. Maybe you did that because of the ref(ref(ref(x))) or similar. Will do some tests with what I think should be correct and see what will happen.

Greetings,
Ilir

Ilir-Liburn:
Hello Mahmoud,

Because I ignored cycles detection, I also ignored fix handling ref(tmp) = NULL situation. Applied and fixed now.
While debugging, I also looked a little bit what cycles detection does, now I have a bit more of the understanding, but wonder does it have to be so complex. Will look further to better understand the issue.

Greetings.
Ilir

MahmoudFayed:
Hello Ilir

>> "scratch that, this happens in Ring2B. It seems I found reference cycle ;)"

It's a nice sample, Thanks for sharing

>> "but wonder does it have to be so complex. Will look further to better understand the issue."

(1) Good article to start with: https://en.wikipedia.org/wiki/Reference_counting

(2) I applied the algorithm described here : https://wiki.c2.com/?ReferenceCountingCanHandleCycles
I did changes to match Ring design (Where the reference count is Zero if we don't have references, but have a variable that own the value and can delete it) 

(3) Another popular algorithm : https://pages.cs.wisc.edu/~cymen/misc/interests/Bacon01Concurrent.pdf

(4) The algorithm in (3) is used by PHP : https://www.php.net/manual/en/features.gc.collecting-cycles.php 

Since Ring assignment operator (=) uses Copy By Value - We are in a better situation compared to other languages, because this reduce references usage, I.e. we create variable and delete them (In the end of the scope) without Reference counting (No increment or decrement is required here)

So using Ref()/Reference() functions is just added for some cases (Like teaching students about Data Structures, Design Patterns, etc.)

We have used Ring for many years without this feature, and I am happy that we can do many things without it. 

Adding this feature is useful, and I hope that it will be used correctly, only when this is necessary.

Greetings,
Mahmoud

MahmoudFayed:
Hello Ilir

>> "but wonder does it have to be so complex. "

You can try new ideas/algorithms using this simulator (Written in Ring itself) : https://github.com/ring-lang/ring/blob/master/language/tests/scripts/general/refsim.ring

Greetings,
Mahmoud

Ilir-Liburn:
Hello Mahmoud,

thanks for sharing the links. I discovered reference cycles 20 years ago. When one German employee drew it on the board and asked me how to solve it. I had no idea at that time and I didn't really need that, it was just academical talk.

Now 20 years later, I discovered Ring which have nice feature: Items structure where we can use pValue to point to another item, and that can be circular (two items pointing to each other for example). One item decreases reference to another item, and then resets pValue to itself. Then it finds there is reference to itself (from another item) and decreases it. Now another item finds that reference to previous item is zero and releases the memory. Then resets pValue to itself and finds that reference is zero. Finally, releases its own memory and cycle is completed.

> It's a nice sample, Thanks for sharing

This sample is showing where the problem lies: reference is increased in advance, therefore on assignment (=NULL) reference is decreased to 0 and memory is not released, causing memory leak.

> You can try new ideas/algorithms

That's exactly what I will do next. Because I don't see why Lists based referencing needs to differ against the Items referencing .

Greetings,
Ilir

MahmoudFayed:
Hello Ilir

>> "That's exactly what I will do next. Because I don't see why Lists based referencing needs to differ against the Items referencing ."

It's about the context and use-cases

Before Ring 1.18 - We uses Items referencing as an *internal* operations by Ring (i.e. Ring VM code - Not Ring Code)
So we control how this will be used, and we are sure that we will not create cycles 

But in Ring 1.18 - We uses Lists referencing because we provide the Ref()/Reference() function - No the developer can use it in Ring Code and this code can create cycles

So Items referencing is just a limited implementation for specific use-cases 
While List referencing is general & complete implementation 

You already invested a lot of time in Ref()/Reference() So I will not explain it, I just will explain Items referencing and Why it has been added

In the beginning, this was outside my plan, Having (=) that uses copy by value, let me think that I will not need this, but then I discovered the next use-case (Maybe 5 years ago or more)

aList = [ 1:10 , 1:10]
mytest( aList, aList[1] )
func mytest aList1, aList2  
      aList1 = NULL # We have the list by reference and we can delete it 
      # Q: What about aList2 ? It's a sublist from aList1 which is deleted! 

So Items referencing is introduced to solve this particular rare case, where we increase the Item reference when we pass it to a function 

Here the function get the parameter: https://github.com/ring-lang/ring/blob/master/language/src/vmfuncs.c#L481
Here we call a function that increment the reference: https://github.com/ring-lang/ring/blob/master/language/src/vmvars.c#L345
Here we have this function: https://github.com/ring-lang/ring/blob/master/language/src/vmgc.c#L44

After that, When Ring added using C pointers by wrapping them inside a Ring list of three items 
To support extensions written in C/C++ like the Files Extensions which contains fopen(), fclose(), etc.

I started to think, what happens if we used fopen() and forget to use fclose() ?
What happens if we copied a pointer then closed the file though one of them
fp = fopen( filename(), "r")
fp2 = fp
fclose(fp)
# Q: what happens to fp2 ?

So In this case too Ring uses Items Referencing, when we use (fp2 = fp) - We increment the reference - But both of them is a weak reference when using fclose() 
i.e. using fclose(fp) we can close the file, set the pointer to NULL - and fp2 will be NULL 
Or using fclose(fp2) we also close the file, set the pointer to NULL - and fp will be NULL 
But if we don't use fclose() - And one of the variables goes out of scope, The other variable will still have access to the File handle

fp2 = NULL
test()
? fp2   # Still have access to the file handle
#  Now fp2 goes out of scope and fclose() will be used automatically by Ring to close the File

func test
	fp = fopen( filename() , "r")
	fp2 = fp
        # Now fp goes out of scope and will be deleted

In summary, Items referencing was just a simple solution to these use-cases. All of these use-cases are under our control (Ring VM code) - But in Lists referencing and Ref()/Reference() - The full control is given to the Ring programmer - So cycles could happen and we detect/remove them

Greetings,
Mahmoud

Ilir-Liburn:
Hello Mahmoud,

I analized ref(tmp)=NULL issue, this goes beyond memory leak. Problem is that ref(tmp) returns self variable followed by the stack assignment going into NULL string assignment, causing decrease of the list reference (first leak), followed by the variable leak finally because we don't have source (tmp) variable here anymore.

There is a workaround, but it is not clean, e.g. might be error prone. We can catch tmp variable from the stack before it is replaced by the self variable and then pass it further to GC where it can replace self variable on the stack (in case newref is set and refcount is 0 during deletion).

I think we should throw an error here saying reference feature can be used only as rhs (right hand side) operation, never lhs (left hand side). It is up to you, maybe you have better idea.

Greetings,
Ilir

Ilir-Liburn:
Hello Mahmoud,

I stand corrected, I made analyses on wrong premise while I was trying to immediately delete self variable pointing to the list.
Here is the correct flow: reference on the list is decreased, therefore variable and list stay intact, but NULL assignment removes list from the variable. This is in a way correct because tmp variable still holds the list, but container variable is not pointing to it anymore. I also forgot to mention that I use do-again loop to test for a memory leak.
Solution to this seems to be simple: instead of getting container variable and deleting it, we should rely on the list and delete it independently of the variable (by deleting them each separately).

I will try this tomorrow and if it works, make a pull request.

Greetings,
Ilir

MahmoudFayed:
Hello Ilir

>> "Solution to this seems to be simple: instead of getting container variable and deleting it, we should rely on the list and delete it independently of the variable (by deleting them each separately)."

Instead of fixing the side-effects, It's better to fix the original problem

In this line of code
ref(tmp) = NULL

The original problem is related to Reference Counting 
Because ref(tmp) increased nReferenceCount by 1

To fix this, In the Assignment operation - We check the (Left Side)
If it's Reference (nReferenceCount  > 1) and (lNewRef Flag is 1, i.e. the reference is not assignment)
Then we know that ref() is used at the Left Side 
The solution could be 
(1) Decrement nReferenceCount  
(2) Set lNewRef Flag to 0

This solution is not tested, but I think it could solve the original problem

Greetings,
Mahmoud

MahmoudFayed:
Hello Ilir

Fixed in this commit: https://github.com/ring-lang/ring/commit/53583d9498157cc865a4425adb159e06a6bb4103

Please test it and be sure that it's working as expected

Greetings,
Mahmoud

Ilir-Liburn:
Hello Mahmoud,

tested, no leak detected. But that doesn't answer the question what is the meaning of ref(tmp) = NULL?

Does the user thinks he is getting reference to the tmp variable (instead of reference to the content of the tmp variable, e.g. the list) and wants to assign NULL value to it? Obviously, it is same (minus reference) like saying

[] = NULL

I still think we should throw an error here saying: list cannot be reassigned while having active references (or similar) because user could also try

ref(tmp) = [1,2,3] // expecting list to be replaced by copy of [1,2,3], etc.

because user could think by using the reference on the left side, he is protecting (threads?) list from deletion or similar.

Greetings,
Ilir

MahmoudFayed:
Hello Ilir

>> "tested, no leak detected. "

Thank you very much

>> "what is the meaning of ref(tmp) = NULL?"

It means 
(1) Delete the reference that is already created by ref() - Because ref() exist in the Left-Side
(2) [] = NULL   (Which doesn't nothing)

>> "I still think we should throw an error here saying: list cannot be reassigned while having active references"

We don't need to do this
Also, List can be reassigned while having active references 

aList = [1,2,3]
aList2 = ref(aList)     # Create a reference    
ref(aList) = [4,5,6]     # Create another reference & delete it  ---> This does nothing
? aList	 		# 1,2,3
aList = [4,5,6]           # List can be reassigned while having active references 
? aList  		        # 4,5,6 (New Value)
? aList2 		        # 1,2,3 (Old Value)

Greetings,
Mahmoud

MahmoudFayed:
Hello Ilir

>> "we again have memory leak! Because list is assigned through assignment pointer, not stack assignment."

Thanks for the report, I will check it

Greetings,
Mahmoud

Ilir-Liburn:
Hello Mahmoud,

You're Welcome. I progressed further supporting ref(tmp) = [1,2,3] and similar, only to come to conclusion that all what I'm doing in case of such statement is: avoiding container damage. Here is what I think about it. Following case

```Ring
tmp = []
ref(tmp) = [1,2,3]
```

Should result in

```Ring
[] = [1,2,3]
```

basically, checking if left is equal to right: means no effect in this case, while

```Ring
tmp = []
? ref(tmp) = [1,2,3]
```
should result in printing false.

Greetings,
Ilir

MahmoudFayed:
Hello Ilir

>> "we again have memory leak! Because list is assigned through assignment pointer, not stack assignment."

Thanks again for the report
Fixed in this commit: https://github.com/ring-lang/ring/commit/54d10debad8953c572fd3ef6e9349cf4dc0ec258

Also this fix take in mind using (new object) on the right side instead of ([1,2,3])
Example:

while true
	test()
end

func test
	tmp = [1:10]
	ref(tmp) = new point
	? refcount(tmp)

class point
	[1,2,3]

Greetings,
Mahmoud

MahmoudFayed:
Hello Ilir

>> "should result in printing false."

Yes, I tested this and it works

tmp = []
? ref(tmp) = [1,2,3]
? ref(tmp) = tmp

Output
0
1

As expected

Greetings,
Mahmoud

Ilir-Liburn:
Hello Mahmoud,

following sample fails

```Ring
do
tmp = new Temp { data="test" }
tmp2 = ref(tmp)
ref(tmp) = [1,2,3]
tmp { data="test2" }
again 1
```

Line 5 Bad parameter type! That is because you are avoiding container damage only in a case of the new ref. I think you are also missing such tests in case of the list items, etc. I will do more testing tomorrow after merging changes.

Greetings,
Ilir

Ilir-Liburn:
Hello Mahmoud,

I finished merging and fixing, all tests are passing, circular references are handled as well.

Minor question: what about referenced list printing? I'm currently using following:

0503E180&2 // pointer reference count 2

helps me identify referenced list and a number of references. 

Major question: what about referenced list by the for-in loop? Such reference should count, otherwise we could get wrong results or crashes. I'm going to prepare test case for this.

Greetings,
Ilir

Ilir-Liburn:
Hello Mahmoud,

Hum, I stumbled to curious test where reference is increasing for each top iteration. Will look into it tomorrow.

```Ring
c = [:wow]
for t=1 to 30
	? c
	a = [1,2,3]
	for e in a
		if e = 1 e = ref(a) ok
		b = [2, ref(a)]
		b[1] = ref(b)
		if e = 2 e = ref(b) ok
		if e = 3 e = ref(c) ok
		? refcount(a)
		? refcount(b)
		? refcount(c)
	next
next
```

Greetings,
Ilir

MahmoudFayed:
Hello Ilir

>> "Line 5 Bad parameter type! That is because you are avoiding container damage only in a case of the new ref. I think you are also missing such tests in case of the list items, etc."

I will check this, Thanks for the report

>> "Minor question: what about referenced list printing? I'm currently using following:"

Nice Idea - Better to use: 0503E180 (RC: 2)  - Please send a pull request

>> "Hum, I stumbled to curious test where reference is increasing for each top iteration. Will look into it tomorrow."

I will check this, Thanks

Greetings,
Mahmoud

Ilir-Liburn:
Hello Mahmoud,

I'm not sure yet about referenced list printing, e.g. do we need to differentiate list and object, for example

[This Attribute Contains A List] (RC: 2)

There is more important thing to think about, here is modified previous sample

```Ring
c = [:wow]
for t=1 to 1
	? c
	a = [1,2,3]
	for e in a
		? "---"
		? a[1]
		? "---"
		if e = 1 e = ref(a) ok
		b = [2, ref(a)]
		b[1] = ref(b)
		if e = 2 e = ref(b) ok
		if e = 3 e = ref(c) ok
		? refcount(a)
		? refcount(b)
		? refcount(c)
	next
next
```
```
wow       
          
---       
1        
---       
3        
2       
1        
---       
04EB0FE0 (RC: 3)
2        
3       
          
---       
3       
3       
1        
---       
0.00      
---       
4       
2        
2       
```
As you can see, last iteration is giving zero number output at a[1] which is clearly wrong. Attempt to print such field using ? ">>"+a[1]+"<<" raises an error for example.

Greetings,
Ilir

MahmoudFayed:
Hello Ilir

Thanks for the report
I will check all of these issues once I return from my vacation in 2023.01.15

Greetings,
Mahmoud

Ilir-Liburn:
Hello Mahmoud,

You're Welcome. While working on the references feature, I will inform you if there is a new issue found.

Greetings,
Ilir

Ilir-Liburn:
Hello Mahmoud,

more info about the reference development, I prepared for-in sample issue:

```Ring
a=[:a]
b=[:b]

insert(a,0,ref(b))
insert(b,0,ref(a))

for y in b
	? y
	? copy("*",10)
	b = []
	? y
next

? :done
```

Ring2B in current state gives

```
0509AF60 (RC: 2)
b

**********
0.00
done
```

Clearly, for-in reference must be added to the list reference count, and I confirmed this while debugging by increasing reference.

Here is what Ring 1.18 in current state gives:

```
b
a

**********
b
a

b
**********
b
done
```

I see here two issues, first is b = [] where for-in loop continues iteration afterwards using the old list, not respecting user wish to  nullify it.

Second happens because insert (add,del?) function(s) do not support references yet.

Greetings,
Ilir

Ilir-Liburn:
Hello Mahmoud,

After adding for-in reference count to list count, I prepared sample to share. I think it is self explanatory, but pay attention to sudden reference count jump from 1 to 3. 

```Ring
a=[:a]
b=[]

for x in a
	? refcount(a)
	b + ref(a)
	for y in b
		? refcount(a)
		? copy("*",10)
		? b
		a = []
	next
	? x
	? refcount(b[1])
	? type(b[1])
	? copy("*",10)
	a + ref(b)
	? a
	b = NULL
	? refcount(a)
	? type(a[1])
next

? :done
```

```
1               
3               
**********      
048AE550 (RC: 3)
                
a               
1               
LIST            
**********      
048B1040 (RC: 2)

1             
LIST      
done
```

Greetings,
Ilir

Ilir-Liburn:
Hello Mahmoud,

while preparing other samples, I came to curious case which I think is a bug:

```Ring
y = new Temp

class Temp
	y = :y
```

```
Line 4 Error (R31) : Trying to destroy the object using the self reference
```

If object is about to be assigned to variable named as object attribute, error is raised. This seems to be related to the scope resolution of the variable. Global scope should be visible only inside object methods, I think following should be disallowed:

```Ring
y = new Temp
? y.z // NULL

class Temp
	z = y
```

Ilir-Liburn:
Hello Mahmoud,

here is a new puzzling issue:

```Ring
y = new Temp
referer(y)

y { t = :t }

z = ref(y)
referer(z)

func referer x
	? refcount(x)
	? type(x)
	? x
	x = 0

class Temp
	t
```

This causes crash on Ring 1.18 and assertion on Ring2B. But more interesting is what happens if we use ref(y) at first call. Then code executes successfully (both 1.18 and 2B), which I do not comprehend right now because I expected ref(y) to pass container to the function and x = 0 to destroy its content (needed later by braces).

Anyhow, I came to conclusion that I shouldn't do anything regarding passing lists to the functions (except list items) because lists are passed by reference in Ring. Hope I came to the right conclusion.

Greetings,
Ilir

MahmoudFayed:
Hello Ilir

>> "This causes crash on Ring 1.18 and assertion on Ring2B."

This error is not releated to ref() usage

This simple test demonstrates that

x = new Temp
x = 0
x { y=10 }

class Temp
	t

I think this will lead to a simple soltion/fix

Greetings,
Mahmoud

MahmoudFayed:
Hello Ilir

>> "This causes crash on Ring 1.18 and assertion on Ring2B."

This error is not related to ref() usage

This simple test demonstrates that

x = new Temp
x = 0
x { y=10 }

class Temp
	t

I think this will lead to a simple solution/fix

Greetings,
Mahmoud

MahmoudFayed:
Hello Ilir

>> "This seems to be related to the scope resolution of the variable. Global scope should be visible only inside object methods"

(1) Access to Global Variables is a feature that provide great power which comes with responsibility
(2) A lot of variables & constants are defined by libraries in the Global Scope - We need to access these things in the class region
(3) The Object take care of itself to avoid destroying itself from the class region & the class method - this protect is enough to catch problems 
(4) Using self.attribute in the class region is enough to avoid conflicts - But in practice I don't use it, Since (Load Package) is enough to load libraries (classes) in another global scope while keeping the constants shared (through using Load without the Package keyword)

Greetings,
Mahmoud

MahmoudFayed:
Hello Ilir

>> "first is b = [] where for-in loop continues iteration afterwards using the old list, not respecting user wish to nullify it."

Normal Behavior - Because using Ref() increases the reference count which means deleting the list will just delete one of the reference and the list will stay because other references exist

To be sure that this is not a (for-in) issue, See this test program

a=[:a,:b]
b=[:b,:c]

for y in b
	? y   # Print Y using Item Reference by For-In
	? copy("*",10)
	b = [] # The List is deleted
	? y     # We still have Item Reference So we can use it 
        # Since b is deleted and the list is empty - The for loop will be terminated 
next

Output:

b
**********
b
done

>> "Second happens because insert (add,del?) function(s) do not support references yet."

Yes, these functions need to be updated/revised to support Ref() usage
Also we need to update printing lists to avoid infinite loops when printing circular references

Thanks for the reports
Keep up the GREAT WORK

Greetings,
Mahmoud

? :done

Ilir-Liburn:
Hello Mahmoud,

> This error is not related to ref() usage

That was discovery after another discovery

```Ring
tmp = []
test(ref(tmp))

func test x
   x = 0
```

here ref() is sending container variable to test function destroying container variable content. Container is not deleted (dontdelete flag), but var value (referenced list) is lost, therefore deletecontainer functionality is damaged afterwards. After this discovery, I tried what happens in general if variable is an object and found that brace object is damaged (not nullified, pointing to some item released at memory pool)

Greetings,
Ilir

```

**Comment by Ilir-Liburn** [comment]
Hello Mahmoud,

> Access to Global Variables is a feature that provide great power which comes with responsibility

This is more question about what is allowed and what is not allowed during an object construction. Normally (in other languages), member variables are allowed to be constructed from extern source through constructor (hence init function), while internally only as constants.
Ring uses its own way which might or not (we will see in Ring2C) cause problems while targeting other languages.

Greetings,
Ilir

**Comment by Ilir-Liburn** [comment]
Hello Mahmoud,

> Normal Behavior - Because using Ref() increases the reference count which means deleting the list will just delete one of the reference and the list will stay because other references exist

The point is: for-in points to old list after deletion, which might be correct (for example in languages implementing iterator), but can also be incorrect (from the user perspective) and not efficient (memory consumption, memory not released: cache hits decreased). Ring2A-B uses later.

Greetings,
Ilir

**Comment by Ilir-Liburn** [comment]
Hello Mahmoud,

> That was discovery after another discovery

I stand corrected. Dontdelete flag prevents variable assignment, therefore container remains unchanged. I don't know in which case I've seen container damage, maybe in some of my previous tests before assignment(s) change. If I find something new, will let you know.

Greetings,
Ilir

**Comment by MahmoudFayed** [comment]
Hello Ilir

>> "If I find something new, will let you know."

Thank you very much
I added all of the issues to my To-Do List
Also, If you fixed one of the reported issues, Please send a pull request

Greetings,
Mahmoud

**Comment by Ilir-Liburn** [comment]
Hello Mahmoud,

You're Welcome. I do not have a pull requests at this time, but I have proposals. First is to change objectid to accept lists in general now that we have

`
0458F100 (RC: 2)
`

so that user can identify lists. Related to that is second proposal: to add additional parameter to the reference function allowing to set id, for example

`
reference(mylist,"mylist")
`

by changing ring_sys_temp variable name and detecting that change in list print so that we have

`
mylist (RC: 2)
`

instead. Does changing the name of the container variable have effect, e.g. do we break something by doing that?

Greetings,
Ilir

**Comment by MahmoudFayed** [comment]
Hello Ilir

>> "First is to change objectid to accept lists in general"

Nice suggestion, Added to my To-Do list

>> "add additional parameter to the reference function allowing to set id"

Doesn't looks natural for me, Sure you can add it as another function or special feature in Ring2B
But if you really need this, It's better to do a search in the local scope of the caller, find the variable name (Looks slower but will give you the same result without changing ref/reference() )

>> "Does changing the name of the container variable have effect, e.g. do we break something by doing that?"

The name is defined as a constant: RING_TEMP_VARIABLE
This constant is used 4 times across 3 files (rinapi.c, vmfuncs.c and vmlists.c)
The usage is just to determine the Name - I don't see checking the name 
So I think changing RING_TEMP_VARIABLE is possible without problems (Not tested)

Greetings,
Mahmoud

**Comment by Ilir-Liburn** [comment]
Hello Mahmoud,

> But if you really need this, It's better to do a search in the local scope of the caller, find the variable name 

No, list can be shared between different variables, or it can originate only from the list item, etc. Right way is to give to user choice of making identifier.

But that's not import now, following sample puzzles me

```Ring
cnd = 0
do
ret = test()
again cnd
? ret
? refcount(ret)
ret { data = NULL }
? ret
? :done

func test

cls = [new Temp { data="test" }]
cls2 = [new Temp { data="test2" }]
cls[1] { data="test1" }
cls2[1] { data="test4" }
return test2(ref(cls2[1]))

func test2 x
	y = test3(x)
	return y

func test3 x
	return ref(x)

class Temp
	data
```

Result should be RC: 1, but I can't get it right. I mean, right without damaging list/container getting attached to the temp variable.

Greetings,
Ilir

MahmoudFayed:
Hello Ilir

Tested using Ring 1.18
I get the next output

data: test4

1
data:

done

I think this is the correct output since RC is 1?

Anyway, Once I return from my vacation I will do more checks, If there is a problem, I will fix it

Greetings,
Mahmoud

MahmoudFayed:
Hello Ilir

>> "First is to change objectid to accept lists in general"

Added in this commit: https://github.com/ring-lang/ring/commit/5327e1d00d7ca934de2136687d0f5206f28a928a

Greetings,
Mahmoud

PR #1494, opened 2023-07-28
What is a race condition?

azzeddine2017:
A race condition occurs when a program depends on the  timing of one or more events to function correctly. A race condition typically occurs when multiple threads access a shared variable (or state) at the same time.


MahmoudFayed:
Thanks for the new sample

PR #1496, opened 2023-08-02
Update Developing the Ring REPLConsole adding CMD commandsringrepl.ring

azzeddine2017:


MahmoudFayed:
Hello Azzeddine

I prefer to keep RingREPL as simple as it
Your idea is nice, But it's better to have it as external package (Using RingPM)

Keep up the GREAT WORK

Greetings,
Mahmoud

PR #1499, opened 2024-04-28
Expanding RingVM htable functionalities

0xNinjaCyclone:
Hello Ring Team

After [this conversation](https://groups.google.com/g/ring-lang/c/AT7M_uh5JwI) with **Ilir Liburn**‏ while facing a problem in returning a hashtable via Ring C API.
I decided to expose the hash table functionalities to the ringworld to facilitate extension developers' work.
So I expanded its functionalities in order to expose them.
My update includes :-
- Expanding the item's data types 
- Adding insert/update/delete/clone/compare/find functions
- Adding functions facilitates traversing the table
- Adding functions exposes some information about the table like `haskey?` and `getkeys`
- Fix issues like the counter not decrementing when an item is deleted and the insert function not checking whether the new key already exists inside the table resulting in table pollution

My updates don't break any old functions. 
I have tested it well on a Linux machine and it works as expected.
I am currently working on the extension, and I already finished it but I wanted to send PRs separately to avoid confusing the reviewers.

Best regards,
**Abdallah Mohamed**.

### DIFF
--- a/extensions/ringsockets/ext/sockets.c
+++ b/extensions/ringsockets/ext/sockets.c
@@ -26,9 +26,9 @@ void ring_vm_socket_init(void *pPointer) {
 
 #ifdef _WIN32
     if (!g_bWinsockInitialized) {
-		RING_API_ERROR("WSAStartup Failed");
-		return;
-	}
+        RING_API_ERROR("WSAStartup Failed");
+        return;
+    }
 #endif
 
     if(RING_API_PARACOUNT < 2) {
@@ -58,7 +58,7 @@ void ring_vm_socket_init(void *pPointer) {
 
 
     sock->addr = NULL;
-	memset(&sock->hints, 0, sizeof(sock->hints));
+    memset(&sock->hints, 0, sizeof(sock->hints));
     sock->sockfd = INVALID_SOCKET;
     sock->hints.ai_family = (int) RING_API_GETNUMBER(1);
     sock->hints.ai_socktype = (int) RING_API_GETNUMBER(2);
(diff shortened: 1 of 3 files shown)


MahmoudFayed:
Hello Abdallah 

I think that these updates should be developed as an extension that can be installed using RingPM
So, I am not interested in integrating these updates with Ring Compiler/VM 

If you need it, you can have it in your fork.

Greetings,
Mahmoud

0xNinjaCyclone:
Hello Mahmoud

I'm currently working on an extension for Ring, this extension has a function that needs to return a hashtable ( That's very common, and future extensions definitely will need that ). 

So from that standpoint, I decided to do it myself after discussing with Ilir to facilitate the development of future extensions, and those functionalities are very useful for the VM itself.

Without expanding the core hashtable functionalities, the extension developers wouldn't be able to deal with it and return it to the high-level. 

The case I'm trying to explain is as follows: An extension needs to return a hashtable, so it uses those functions and returns a managed pointer. On the other hand, the high-level developers interact with it using an embedded extension with the VM like `list_e` or a separate extension, that extension also relies on the hashtable core functionalities.

So I ask you to review your decision again or tell us an appropriate solution for this problem.

Thanks.

MahmoudFayed:
Hello Abdallah

Don't update Ring VM HashTable implementation (It's designed to be a small implementation for specific purpose, i.e. to be used by Ring Compiler/VM to manage tables for functions/classes) - Being small implementation is a design goal. 

Solution: Just keep all of the code that you need in your extension folder, develop the extension as an external extension in your GitHub account, once finished create a package for your extension then we can add it to RingPM registry: https://github.com/ring-lang/ring/tree/master/tools/ringpm/registry

Greetings,
Mahmoud

PR #1500, opened 2024-08-18
Ring syntax highlighting for nano

ysdragon:
Ring syntax highlighting for nano

https://github.com/ysdragon/ring-nano


MahmoudFayed:
Thanks for your contribution

PR #1502, opened 2024-08-19
Update performance to clock in the Command Line Options chapter.

ysdragon:


MahmoudFayed:
Thanks for the update

PR #1503, opened 2024-08-20
Add other Linux systems download guide 

Programmer-Om:
# Added other Linux systems guide 

- Added : Fedore , CentOS , OpenSESU , Debian , Arch 
- Updated main README.md file 
- Changed README_UbuntuLinux.md to README_Linux.md
- Download guide review


MahmoudFayed:
At the current stage, we provide formal support for Ubuntu Linux (Only)
For using Ring on other distribution, it's the community choice/work
We can't maintain/support this for now (This requires updating files that install dependency, test each new release on all of these distributions).

PR #1505, opened 2024-08-21
Added missing `libuv1-dev` and added build-essential to `installdepubuntu.sh` script

ysdragon:
Added missing libuv1-dev and added build-essential as it includes (gcc,g++,make).


MahmoudFayed:
Thanks for the update

PR #1510, opened 2024-08-23
Create info.ring

Programmer-Om:
Hello Mr.Mahmoud . I want to ask you about creating info command . 
  It will give info about : 
- Ring version . 
- Ring Package Manager version ( if you provided it for me because i think that is there no information about it's version) .
- Some information about Ring and Ring Package Manager ( like license , copyright  )
  My greetings , Programmer-Om .

### DIFF
--- a/tools/ringpm/commands/info.ring
+++ b/tools/ringpm/commands/info.ring
@@ -1,0 +1,6 @@
+/*
+	  Title :	The Ring Package Manager 
+	  Date  : n/n/n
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Hello
We don't need this
RingPM version is the same as Ring version
When we run ring/ring2exe/ringpm/ringrepl/etc. 
We display the version on the screen
Also, Ring provides the version() function

PR #1517, opened 2024-09-04
Fix lib path when distributing the application for Linux with specific libraries

ysdragon:
This may also apply to macOS.


MahmoudFayed:
Thanks for the update
Yes, this apply for macOS too (I will modify it)

PR #1520, opened 2024-09-10
Refactor the build function in `build/buildgcc.sh` and make the script callable from any directory

ysdragon:


MahmoudFayed:
Thank you very much for improving the script file

PR #1521, opened 2024-09-13
Print the version we are updating from and the newer version in RingPM

ysdragon:
Screenshot
ringpm

### DIFF
--- a/tools/ringpm/commands/install.ring
+++ b/tools/ringpm/commands/install.ring
@@ -87,7 +87,7 @@ class InstallCommand
 				ok
 			ok
 		cPackageFolderName = GetPackageFolderName(aPackageInfo)
-		see " ---> " + cPackageFolderName + nl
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
This is nice, Thanks for the update

ysdragon:
Thank you for accepting the pull request. 
I have a suggestion for RingPM: when a user interrupts the download process or when their connection fails, it should switch back to the previous version in `package.ring` and `lock.ring` so they can update again.

MahmoudFayed:
You are welcome

I prefer the current behavior which doesn't add complexity for low benefit

If we applied this suggestion, this will not solve the problem because the package files could be a mix of old and new files. Some packages have scripts and zip files and could include complex procedures

With respect to the current behavior
if a problem happened during installing/updating a package the recommended fix is

1- remove the package
2- install the package again

package could be removed using ringpm remove
Or by deleting package folder inside ringpm/packages

Also, we can change the package version inside ringpm/packages to an old version, it just a simple Ring script

PR #1522, opened 2024-09-16
CMake and Ring Website

ysdragon:
Hello Mahmoud,

Thank you for adding me on the Ring Website, I appreciate that very much.

The first commit is to get the Distro Name from `/etc/os-release` instead of `lsb_release` because it's not installed by default on some distros, even though it is installed by default in Ubuntu, Void Linux, for example.

The second commit is to fix a letter in my name.

Thank you once again for adding me on the Ring Website.

Best regards,
Youssef


MahmoudFayed:

PR #1523, opened 2024-09-24
Update `buildgccstatic.sh` - Make the Ring executable fully static

ysdragon:
This update resolves issues where the `ring` executable fails to run due to missing `glibc` versions on older systems (specifically when running the executable compiled with a newer glibc version).

```c
./ring: /lib/aarch64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by ./ring)
./ring: /lib/aarch64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ./ring)
./ring: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./ring)
./ring: /lib/aarch64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./ring)
```
This may also apply to the ring2exe `-static` argument on Linux too (If the ring script doesn't require any extensions, of course).

### DIFF
--- a/language/build/buildgccstatic.sh
+++ b/language/build/buildgccstatic.sh
@@ -1,7 +1,7 @@
 cd ../src
 
 clear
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Thanks for the update

PR #1525, opened 2024-10-13
Add Ring/Ring2EXE build system for VSCode

ysdragon:
I have implemented Ring and Ring2EXE build system for VSCode.
image


MahmoudFayed:
Thanks for the update

PR #1527, opened 2024-10-28
Update fastpro.c - Fix compilation under Linux

ysdragon:


MahmoudFayed:
Thanks for the update

PR #1529, opened 2024-11-04
Fix fastpro build scripts

ysdragon:
The `buildgcc.sh` and `buildclang.sh` scripts in RingFastPro do not output the library to the lib directory.


MahmoudFayed:
Thanks for the update

PR #1532, opened 2025-02-14
Add missing `endif` keyword for the VS Code extension

ysdragon:


MahmoudFayed:
Thanks for the update

PR #1536, opened 2025-05-07
Fix LibUV compilation on Linux with GCC 14/Clang 19+

ysdragon:
* Add explicit casts to `uv_mutex_lock`, `uv_mutex_unlock`, and `uv_mutex_destroy` when passing them to ring_vm_mutexfunctions.
This resolves incompatible function pointer type errors caused by stricter type checking in newer compilers.

* Updated `ring_libuv.c` by regenerating it.

While testing LibUV on Linux, the first example (`ex1.ring`) works; however, the remaining examples do not function as they do on Windows. They are hanging, seemingly in an infinite loop.


MahmoudFayed:

PR #1544, opened 2025-06-19
Update language/CMakeLists.txt - Require version 3.16 and improve OS detection messages

ysdragon:
- Fixed CMake Deprecation warning.
- Fixed `CMP0177 policy` warnings caused by using absolute parent directory path instead of using relative paths.
- Now, it shouldn't error on Termux.


MahmoudFayed:
Thanks for the update

PR #1546, opened 2025-06-30
Refactor(scanner): Use data-driven tables for operator parsing

idrassi:
Refactors the operator scanning logic in `ring_scanner_readchar` to be data-driven instead of using a large chain of if/else statements.

Two static arrays, `OP_MULTI` and `OP_COMPOUND`, are introduced to define multi-character operators (e.g. `++`, `<<`) and compound assignment operators (e.g. `+=`, `*=`). The parsing logic now iterates through these tables which simplifies the code and improves readability.

```
 The Report Summary
 Tests Count : 800
 PASS        : 800
 FAIL        : 0
```

### DIFF
--- a/language/src/scanner.c
+++ b/language/src/scanner.c
@@ -2,45 +2,38 @@
 
 #include "ring.h"
 
-/* Data Compound Operators */
-typedef struct CompoundOperator {
-    const char *cOperator;
-    const char *cCompound;
-    int nToken;
-} CompoundOperator ;
+/* Data Operators (Compound and Multi-character) */
+typedef struct OperatorInfo {
+	const char *cOperator;
+	const char *cSecond;
+	int nToken;
+} OperatorInfo ;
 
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Hello Mounir

Thanks for the update, Yes, all the tests pass, and the code size is reduced by over 40 LOC
Also, this update reduces the binary file (ring.dll) by 512 bytes.
When I compile PWCT2 after this update, I get the same performance.

Just one little comments:
This struct
typedef struct CompoundOperator {
    const char *cOperator;
    const char *cCompound;
    int nToken;
} CompoundOperator ;

Is the same as this struct
typedef struct MultiCharOperator {
    const char *cOperator;
    const char *cMulti;
    int nToken;
} MultiCharOperator ;

They contain the same members, so one of them is enough.

MahmoudFayed:
Hello Mounir

You are welcome :D 

Greetings,
Mahmoud

PR #1547, opened 2025-07-01
Scanner: Fix bug in uninitialized variable errors message when name contains underscores

idrassi:
When an identifier containing underscores was used without being initialized, the interpreter would report a runtime error with a mangled variable name where all underscores except a potential trailing one were stripped. This causes confusion since no such mangled name exists in the Ring code.

Unfortunately, the expected output of the test `general/test106.ring` was modified to hide this bug.

The root cause was that the `ring_scanner_isnumber` function performed destructive, in-place modifications on the token string while attempting to validate it. If the token was ultimately determined to be an identifier and not a number, its string content had already been permanently altered.

This patch resolves the issue by refactoring the token classification logic to separate validation from normalization:

1.  The '`ring_scanner_isnumber`' function is now a pure, non-destructive validation function. It correctly determines if a string represents a number without modifying the input buffer.

2.  A new function, `ring_scanner_normalize_number_token`, is introduced. This function is responsible for removing underscores and float suffixes from tokens that have already been confirmed to be numbers.

3.  The `ring_scanner_checktoken` function is updated to first validate a token with isnumber. Only if it is a number is the new `normalize_number_token` function called. This ensures that identifiers are never modified.

The test `general/test106.ring` expected output has been updated to assert the correct, non-mangled error message, preventing future regressions.

Remark: `ring_scanner_isnumber` now takes String pointer directly and doesn't use `strlen` since `pScanner->pActiveToken` length as String will always be the same as the 'char*' length.

```
 The Report Summary
 Tests Count : 800
 PASS        : 800
 FAIL        : 0
```


MahmoudFayed:
Hello Mounir

Thank you very much for the bug report/fix.

Could you please update the pull request to use the latest version of scanner.c and scanner.h

Greetings,
Mahmoud

idrassi:
I have solved the conflicts on the modified files.

There is an extra commit after the merge to revert some changes on unrelated files (they are part on other ongoing changes for which I will create separate PR once ok).

PR #1548, opened 2025-07-02
refactor(rhtable): Correct and optimize hashtable rebuild process

idrassi:
The previous implementation of `ring_hashtable_rebuild_gc` had two flaws: a memory management bug and a major performance bottleneck.

The Bug:
It incorrectly called the non-GC-aware wrappers (`ring_hashtable_newnumber` and `ring_hashtable_newpointer`) instead of their `_gc` counterparts. This caused new hash items to be allocated using the standard system `malloc`, completely bypassing Ring's memory pool manager during a memory-intensive operation.

The Performance Issue:
The logic was inefficient. For every item in the hash table, it would:
1. Free the `HashItem` struct.
2. Free the key string.
3. Re-allocate a new `HashItem` struct.
4. Re-allocate and copy the key string via `strdup`. This resulted in excessive memory churn.
5. `ring_hashtable_rebuild_gc` is called recursively

This commit completely refactors `ring_hashtable_rebuild_gc` to resolve these issues. The new implementation moves existing `HashItem` nodes directly from the old table's buckets to the new, larger table's buckets. It simply re-links the pointers without any deallocation or reallocation of the items themselves.

```
 The Report Summary
 Tests Count : 801
 PASS        : 801
 FAIL        : 0
```

### DIFF
--- a/language/src/rhtable.c
+++ b/language/src/rhtable.c
@@ -158,35 +158,51 @@ HashTable * ring_hashtable_delete_gc ( void *pRingState,HashTable *pHashTable )
 
 void ring_hashtable_rebuild_gc ( void *pRingState,HashTable *pHashTable )
 {
-	HashItem **pArray  ;
-	int nLinkedLists,x  ;
-	HashItem *pItem,*pItem2  ;
+	HashItem **pOldArray  ;
+	int nOldLinkedLists, x, nNewIndex  ;
+	HashItem *pItem, *pNextItem  ;
+
+	/* If the number of items doesn't meet the rebuild threshold, do nothing. */
 	if ( pHashTable->nItems != pHashTable->nRebuildSize ) {
 		return ;
 	}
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Hello Mounir

I have already tested it using Ring Tests, Samples & Applicaitons.

Greetings,
Mahmoud

idrassi:
Thank you Mahmoud for the extra tests and validation.
I'm studying the internals of Ring and I'm glad to be able to enhance the codebase a little bit. Still so many things to learn!

MahmoudFayed:
Hello Mounir

You are welcome :D 
After doing more tests, I discovered that this update break Ring Notepad at startup
So, I restored the original implementation for now.

Greetings,
Mahmoud

MahmoudFayed:
Hello Mounir

Greetings,
Mahmoud

PR #1549, opened 2025-07-04
Update RingPM - Registry - Add: toml extension

ysdragon:
I also added it to the official TOML implementations section: https://github.com/toml-lang/toml/wiki#implementations

image

Thanks.


MahmoudFayed:
Thank you very much for developing the extension and the RingPM registry update

PR #1552, opened 2025-07-06
feat(core): Overhaul string implementation for performance gains

idrassi:
This commit completely refactors the internal `String` implementation to use more modern approach for C data structures. The previous implementation suffered from performance issues due to excessive reallocations and inefficient algorithms.

The key changes include:

- **Redesigned `String` Struct:** The struct now includes an `nCapacity` field to explicitly track allocated memory, distinguishing it from the string's current `nSize` (length). This is fundamental to all other optimizations.

- **Geometric Growth Strategy:** `ring_string_add2_gc` now uses a doubling strategy for capacity when reallocating, similar to std::string in C++ and ArrayList in Java. This changes string building from a quadratic O(n^2) operation to an amortized linear O(n) operation, dramatically reducing system calls for appends.

- **Allocation-Free Case-Insensitive Search:** `ring_string_find3_gc` is rewritten to be allocation-free. It now performs case-insensitive comparisons on-the-fly, avoiding the overhead of allocating and lowercasing two temporary strings for every search.

- **Optimized Integer-to-String Conversion:** `ring_string_setfromint_gc` now uses `snprintf(NULL, 0, ...)` to calculate the required size. If the existing capacity is sufficient, it writes directly into the string's buffer, avoiding a re-allocation cycle.

- **Simplified Memory Logic:** The `ring_string_alloc/free/realloc_gc` helpers have been removed. The logic is now handled cleanly and directly within the primary string functions, making the Small String Optimization behavior more explicit and robust.

- **Remove Unused Buggy Function:** The function  `ring_string_swaptwostrings` was removed because it is unused and it contains a bug with regards to small strings since it was copying pointer address that becomes invalid afterwards.

On my machine, I see a 10% to 20% performance gain on programs that manipulate a lot of strings.

Test suite runs successfully after this change. Ring Notepad and other RingQT applications were also tested successfully.
```
 The Report Summary
 Tests Count : 806
 PASS        : 806
 FAIL        : 0
```

### DIFF
--- a/language/include/rstring.h
+++ b/language/include/rstring.h
@@ -10,8 +10,8 @@
 	#endif
 	typedef struct String {
 		char *cStr  ;
-		unsigned int nSize  ;
-		/* Note : nSize is (Allocation Size - 1) */
+		unsigned int nSize  ; /* Current length of the string, excluding the null terminator. */
+		unsigned int nCapacity  ; /* Allocated buffer size. For SSO, this is RING_STRING_ARRAYSIZE. */
 		char cStrArray[RING_STRING_ARRAYSIZE]  ;
 	} String ;
 	#define ring_string_tolower(pString) ring_string_lower(pString->cStr)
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Hello Mounir

Thank you very much for this wonderful update 

Little comments:

(1) After these updates, the performance of Ring code generator (RingQt extension) is improved by 24% on my machine. 

(2) For testing such updates, (Ring code generator for extensions like RingQt) and (Arabic Poetry Analysis) are more suitable since they have heavy usage for string functions.
  
(3) Similar pull requests should be multiple pull requests so we can revise/test/decide based on each update alone.
Also, this enables us to test the performance improvements based on each change (alone) before mixing them together.

Greetings,
Mahmoud

idrassi:
Hi,

Thank you Mahmoud for the feedback and thank you Mansour for the kind words.

Indeed, code generator is a good benchmark. Your results align with my own tests. I will use RingQT code generator for future benchmarks.
I agree that small PRs are better. This has always been my approach. Just in this case, everything was based on String structure redesign and it would have needed more work to create separate PRs that are stable enough without the others. That being said, I agree with your point 100%. I will do better next time.

PR #1553, opened 2025-07-19
Remove manual free() on curl-managed memory to prevent crash

ysdragon:
According to the official libcurl documentation, memory returned by `curl_easy_getinfo` is owned by the cURL handle and must not be freed by the caller. It is automatically released when `curl_easy_cleanup` is called.

This commit removes the erroneous `free(sValue)` call, which resolves the crash and ensures correct memory management.

Ref: https://curl.se/libcurl/c/curl_easy_getinfo.html


MahmoudFayed:

PR #1554, opened 2025-07-25
Add method to access raw request body

ysdragon:
Previously, the library provided convenient access to URL parameters and multipart form data (e.g., `oServer["param"]`). However, there was no direct way to retrieve the raw request body. This made it difficult to handle common API patterns where data is sent as a single payload, such as a `POST` request with a JSON or XML body.

Developers attempting to process a `curl` request like the one below would find it impossible to get the JSON data:
```bash
curl -X POST -H "Content-Type: application/json" -d '{"message":"hello"}' http://localhost:8080/api
```

#### **Solution**

This PR exposes the `body` property from the underlying C++ `httplib::Request` object to the Ring wrapper.

A new method, `body()`, has been added to the `Request` class. Developers can now access the full, raw request body as a string.

### DIFF
--- a/documents/source/httplib.txt
+++ b/documents/source/httplib.txt
@@ -32,6 +32,7 @@ Server Class Methods
 * cookies() -> aList
 * getFileContent(cFile) -> cString
 * getFileName(cFile) -> cString
+* request().body() -> cString
 
 .. index:: 
 	pair: Using HTTPLib; Example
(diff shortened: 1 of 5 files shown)


MahmoudFayed:
Hello Youssef

Very nice, Thank you very much for the update

Greetings,
Mahmoud

PR #1558, opened 2025-07-30
Update LibCurl binaries

ysdragon:
These binaries are now built with Schannel.


MahmoudFayed:
Hello Youssef

Greetings,
Mahmoud

PR #1568, opened 2025-08-28
Update documents/pygments_lexer - Support function calls

ysdragon:
The RingLexer now supports function calls.


MahmoudFayed:
Hello Youssef

Thanks for the update

Greetings,
Mahmoud

PR #1577, opened 2025-09-25
Vm: Fix unused inner loop in ring_vm_afterscopeidoverflow

idrassi:
The function contained an unnecessary inner loop over y that was not used, causing the switch statement to execute 2 times per bytecode instruction without purpose. This wasted CPU cycles. The fix removes the inner loop, ensuring the switch runs only once per instruction for better performance and correctness.


MahmoudFayed:
Hello Mounir

Greetings,
Mahmoud

PR #1578, opened 2025-09-25
Tests: fix wrong expected output of general_fread test by removing erroneous new lines

idrassi:
The expected output of general\fread.ring test was wrong because it had erroneous new lines added to it. This made the test appear as failing although it was succeeding. This commit removes the erroneous new lines so that the output matches the content of original fread.ring file as stored in git repository (fread.ring is stored in the repository with Unix line ending and UTF-8 encoding).


MahmoudFayed:
Hello Mounir

Greetings,
Mahmoud

PR #1579, opened 2025-09-27
perf(hashtable): replace modulo with bitwise AND for bucket indexing

idrassi:
Replace `nIndex % nLinkedLists` with `nIndex & (nLinkedLists - 1)` in `ring_hashtable_hashkey_gc`. `nLinkedLists` starts at 2 and always doubles, so it remains a power of two. The bitwise mask avoids a division instruction and offers a small performance improvement. It doesn't introduce any functional changes.


MahmoudFayed:
Hello Mounir

Greetings,
Mahmoud

PR #1580, opened 2025-09-28
list: remove per-recursion string allocation in ring_list_sortstr_gc

idrassi:
Replace temporary pivot string allocation with const char * pointing to existing data. Swapping only exchanges Item pointers, so underlying string storage is stable. Eliminates O(log(n)) transient allocations during quicksort recursion, reducing GC pressure and improving performance without changing behavior.


MahmoudFayed:
Hello Mounir

Thank you very much for the update :D

Greetings,
Mahmoud

PR #1624, opened 2026-05-07
Update language/src/genlib_e.c - list2str() newline guard uses nStart

cuiweixie:
## Summary

In `ring_vm_generallib_list2str()`, newline separators between list items were skipped using `x != 1`. The function iterates `x` from `nStart` to `nMax`, so when `nStart > 1` the first emitted item incorrectly received a leading newline.

## Change

Use `x != nStart` for both string and number branches so the first **included** index never gets a prefix newline.

## Note

Local build artifacts were not committed.


MahmoudFayed:
Thanks for the update

PR #1625, opened 2026-05-07
Update language - ring_state_runobjectfile() returns status for ring_state_execute()

cuiweixie:
## Summary

`ring_state_execute()` uses `lCont` (and `exit(RING_EXIT_FAIL)`) for both source and object files. `ring_state_runfile()` already returns an `int` status; `ring_state_runobjectfile()` was `void`, so failures when loading or running a `.ringo` file did not surface as a non-zero process exit.

## Changes

- `ring_state_runobjectfile()`: return `RING_FALSE` on path-too-long or failed `ring_objfile_readfile()`; return `RING_TRUE` after a successful load and VM run.
- `language/include/state.h`: declaration updated to match.

## Note

Embedding docs that still say `void` for this API were not updated in this PR. Local build artifacts were not committed.

### DIFF
--- a/language/include/state.h
+++ b/language/include/state.h
@@ -130,7 +130,7 @@ RING_API int ring_state_runfile(RingState *pRingState, char *cFileName);
 
 RING_API int ring_state_runstring(RingState *pRingState, char *cString);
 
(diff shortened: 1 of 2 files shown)


MahmoudFayed:
Thanks for the update

PR #1626, opened 2026-05-07
Update language/src/genlib_e.c - lines() count for empty string and trailing newline

cuiweixie:
## Summary

Adjust `ring_vm_generallib_lines()` when the argument is a string:

- **Empty string** (`nSize == 0`): return **0** lines (previously returned 1).
- **Trailing newline**: do not increment the line count for a `\n` at the last index, so a string ending with one newline matches “lines separated by \n” without an implicit empty trailing line.

## Files

- `language/src/genlib_e.c` only (no build artifacts).


MahmoudFayed:
Thanks for the update

Note: using nCount = (nSize > 0) is enough (i.e. no need for ? 1 : 0 
I will do this little change myself

PR #1627, opened 2026-05-08
Update language/src/ringapi.c - getcpointer2pointer() typed null handling

cuiweixie:
## Summary

In `ring_vm_api_getcpointer2pointer()`, when the list’s native pointer is `NULL`, the VM only skipped the error path if `RING_CPOINTER_TYPE` was the literal `"NULLPOINTER"`. Extensions that keep the real type name on a null optional pointer were still hitting `RING_API_NULLPOINTER`.

## Change

Also return `NULL` without error when the stored type string equals the requested `cType` (typed null), same outcome as the `NULLPOINTER` sentinel.

## Scope

- `language/src/ringapi.c` only; build artifacts not committed.

### DIFF
--- a/language/src/ringapi.c
+++ b/language/src/ringapi.c
@@ -374,7 +374,8 @@ RING_API void *ring_vm_api_getcpointer2pointer(void *pPointer, int nPara, const 
 				RING_API_ERROR(RING_API_BADPARATYPE);
 				return NULL;
 			} else {
-				if (strcmp(ring_list_getstring(pList, RING_CPOINTER_TYPE), "NULLPOINTER") == 0) {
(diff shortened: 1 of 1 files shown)


MahmoudFayed:
Thanks for your update/suggestion

In reality the current behavior is correct and no change is required

When the pointer is NULL, and the RING_CPOINTER_TYPE is the literal "NULLPOINTER" then we pass the error and return NULL because we use this feature to pass NULL to extensions

If we have another type name (i.e. not "NULLPOINTER"), then this means we are trying to use a C pointer after it's deleted and in this case the function must produce RING_API_NULLPOINTER.

PR #1635, opened 2026-05-19
Update SQLite Chapter

ysdragon:
Hello Mahmoud,

The output block for the first example was incorrectly added after the second example. I've moved it to the correct spot in this PR.

Best regards,
Youssef


MahmoudFayed:
Hello Youssef

Thanks for the revision and update

Greetings,
Mahmoud

PR #1640, opened 2026-08-03
Make the RING_VM_* build flags settable from the build

mayouni:
Hello Mahmoud,

A small build-system report. As with the last one, please treat the diff as
illustration rather than something to merge — I know the C is generated from
PWCT.

## What happens

`language/include/ext.h` offers six switches for the extension modules:

```c
#define RING_VM_LISTFUNCS 1
#define RING_VM_MATH 1
#define RING_VM_FILE 1
#define RING_VM_OS 1
#define RING_VM_REFMETA 1
#define RING_VM_INFO 1
```

They read as build-time options, and `ring.h` already honours them — it
conditions each `#include "math_e.h"` and so on. But because the defines are
unconditional, a value passed from the build is overridden as soon as `ext.h`
is included:

```
gcc -DRING_VM_MATH=0 ...     # no effect: the flag is 1 regardless
```

So the option cannot be exercised. `RING_VM_DLL` immediately below is the
exception, and shows the intended shape — it is already conditional, on
`RING_NODLL`.

## A second, related point

Fixing only the above makes setting a flag to `0` fail to build: `ring.h` stops
declaring the module's prototypes, but the module's `.c` file is still compiled
and references them.

```
vminfo_e.c:6:40: error: use of undeclared identifier 'ring_vm_info_ringvmfileslist'
```

Each module therefore also needs to compile to nothing when its flag is off —
the same condition `ring.h` already applies to the header include.

## What this PR does

Two mechanical edits, no behaviour change:

1. `ext.h` — the six defines wrapped in `#ifndef`, exactly as `RING_VM_DLL` is
   already conditional.
2. `math_e.c`, `file_e.c`, `os_e.c`, `list_e.c`, `meta_e.c`, `vminfo_e.c` and
   `dll_e.c` — each body wrapped in `#if RING_VM_<MODULE>` … `#endif`, after
   the `#include "ring.h"` line.

Point 2 also settles an existing case: with `RING_NODLL` set, `ext.h` defines
`RING_VM_DLL 0` and `ring.h` drops `dll_e.h`, yet `dll_e.c` is still compiled.
Builds get away with it today only by removing that file from the source list
by hand.

## Verified

Against `master` (98914829), compiled to `wasm32-wasi` with `zig cc`:

- **With the flags at their defaults, the output binary is byte-identical** —
  same SHA-256 as the build before this change. Nothing moves for anyone who
  does not set a flag.
- **With `-DRING_VM_REFMETA=0 -DRING_VM_INFO=0 -DRING_VM_MATH=0` and no other
  change** — every source file still in the build list — it compiles cleanly and
  the binary goes from 363,218 to 327,210 bytes (−36 KB, −9.9%).

Being able to leave out reflection or the file layer matters most for embedded
and WebAssembly targets, where the whole VM ships to the user on every page
load. This came up while building [RingScript](https://github.com/mayouni/ringscript),
which runs the Ring VM in the browser.

Thank you for Ring,
Mansour

### DIFF
--- a/language/include/ext.h
+++ b/language/include/ext.h
@@ -2,12 +2,24 @@
 
 #ifndef ringext_h
 #define ringext_h
-#define RING_VM_LISTFUNCS 1
-#define RING_VM_MATH 1
-#define RING_VM_FILE 1
-#define RING_VM_OS 1
-#define RING_VM_REFMETA 1
-#define RING_VM_INFO 1
+#ifndef RING_VM_LISTFUNCS
+	#define RING_VM_LISTFUNCS 1
+#endif
+#ifndef RING_VM_MATH
+	#define RING_VM_MATH 1
+#endif
+#ifndef RING_VM_FILE
+	#define RING_VM_FILE 1
+#endif
+#ifndef RING_VM_OS
+	#define RING_VM_OS 1
+#endif
+#ifndef RING_VM_REFMETA
+	#define RING_VM_REFMETA 1
+#endif
+#ifndef RING_VM_INFO
+	#define RING_VM_INFO 1
+#endif
 #if RING_NODLL
 	#define RING_VM_DLL 0
 #else
--- a/language/src/dll_e.c
+++ b/language/src/dll_e.c
@@ -2,6 +2,8 @@
 
 #include "ring.h"
 
+#if RING_VM_DLL
+
 void ring_vm_dll_loadfunctions(RingState *pRingState) {
 	RING_API_REGISTER("loadlib", ring_vm_dll_loadlib);
 	RING_API_REGISTER("closelib", ring_vm_dll_closelib);
@@ -105,3 +107,5 @@ void ring_vm_dll_closealllibs(void *pPointer) {
 	}
 	ring_list_deleteallitems_gc(pVM->pRingState, pVM->pCLibraries);
 }
+
+#endif
--- a/language/src/file_e.c
+++ b/language/src/file_e.c
@@ -2,6 +2,8 @@
 
 #include "ring.h"
 
+#if RING_VM_FILE
+
 void ring_vm_file_loadfunctions(RingState *pRingState) {
 	RING_API_REGISTER("fopen", ring_vm_file_fopen);
 	RING_API_REGISTER("fclose", ring_vm_file_fclose);
@@ -918,4 +920,6 @@ void ring_vm_file_tempname(void *pPointer) {
 	RING_API_RETSTRING(_tmpfile);
 	#endif
 }
+#endif
+
 #endif
--- a/language/src/list_e.c
+++ b/language/src/list_e.c
@@ -2,6 +2,8 @@
 
 #include "ring.h"
 
+#if RING_VM_LISTFUNCS
+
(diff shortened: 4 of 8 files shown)


ysdragon:
Hello @mayouni,

Could you check my workaround in [`ring-lang-rs`](https://github.com/ysdragon/ring-lang-rs/blob/master/src/extension.rs) With this approach, you don't need to modify the Ring source code at all.

I'm already using it in Ring-Slint to distribute apps for Android and iOS, and I think you can apply the same idea to Zig.

Best regards,
Youssef

mayouni:
Thank you Youssef!
I'll check, but not alone ;)
Since I rely on Claude AI for all this low level staff.
I'll let you know.

mayouni:
Thank you @ysdragon — your approach is better than this PR, and I am closing it
in favour of it.

For anyone finding this later, the technique in
[`ring-lang-rs/src/extension.rs`](https://github.com/ysdragon/ring-lang-rs/blob/master/src/extension.rs)
is: `ext.c` exists only to call the seven `ring_vm_*_loadfunctions()`, so you
drop `ext.c` from the build, provide your own `ring_vm_extension()` calling
just the modules you want, and leave the others uncompiled. One function is
the whole seam, and the Ring source is untouched.

It ports to Zig unchanged. In RingScript (Ring 1.27 → `wasm32-wasi` via
`zig cc`), dropping `ext.c` plus `math_e.c`, `meta_e.c` and `vminfo_e.c` and
supplying my own `ring_vm_extension`:

```
364,226 → 327,354 bytes   (-36 KB, -10.1%)
```

— the same saving this PR was chasing, with no change to Ring at all. So the
`RING_VM_*` defines being unconditional stops mattering: nobody needs them
fixed to select a module set.

One result worth passing on, since it surprised me. Trimming that far broke
every test with `Calling Function without definition: isfunction` — my own
auto-main shim calls `isfunction()`, which lives in `meta_e.c`. Putting
reflection back, three more failed on `floor`, in `math_e.c`, used by the
JSON codec I embed. For an embedder, `refmeta` and `math` can be load-bearing
even when the user's own Ring code never touches them. What was actually
droppable in my case was `vminfo_e.c` alone:

```
364,226 → 357,897 bytes   (-6 KB, -1.7%)   all tests still passing
```

Worth measuring before assuming a module is optional.

@MahmoudFayed — no action needed here. The only thing left is cosmetic: the six
`#define RING_VM_* 1` in `ext.h` read as build switches but cannot be set from
the build, since they are unconditional (`RING_VM_DLL` just below them is the
exception, and shows the shape). If you are ever in that file, `#ifndef` around
them would make them behave as they read. Not worth a change on its own.

Thank you both.

MahmoudFayed:
Hello Mansour & Youssef

You are welcome

Note: Other projects that you could check is Ring for Microcontroller
And the idea of using rconfig.h file: https://github.com/ring-lang/ring/blob/master/extensions/microcontroller/ringpico/ring/include/rconfig.h

This file is used by Ring source code and it's empty in the default Ring version (desktop version)
So, you can customize some Ring choices/decision using it.

Greetings,
Mahmoud

